PR #21, opened 2019-09-20
Update to lifecycle 0.4.0

hone:
This upgrades to lifecycle `0.4.0` which introduces a breaking change as well uses the `api` key in `builder.toml`. This includes updates to the heroku buildpacks + shim to make sure it works.

### DIFF
--- a/builder.toml
+++ b/builder.toml
@@ -10,9 +10,9 @@ uri = "lifecycle.tgz"
   id = "heroku/java"
   uri = "https://github.com/heroku/java-buildpack/releases/download/v0.14/java-buildpack-v0.14.tgz"
 
-# [[buildpacks]]
-#   id = "heroku/ruby"
-#   uri = "https://github.com/heroku/heroku-buildpack-ruby/releases/download/cnb-alpha/ruby-buildpack-vCNB-alpha.tgz"
+ [[buildpacks]]
+   id = "heroku/ruby"
+   uri = "https://github.com/heroku/heroku-buildpack-ruby/releases/download/cnb-alpha/ruby-buildpack-vCNB-alpha.tgz"
 
 [[buildpacks]]
   id = "heroku/procfile"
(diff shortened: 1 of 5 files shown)


hone:
I also updated `heroku/pack-runner` to use `pack` `0.4.1`.

builder.toml
```diff
@@ -4,19 +4,19 @@ build-image = "heroku/pack:18-build"
...
   id = "heroku/java"
-  uri = "https://github.com/heroku/java-buildpack/releases/download/v0.13/java-buildpack-v0.13.tgz"
+  uri = "https://github.com/heroku/java-buildpack/releases/download/v0.14/java-buildpack-v0.14.tgz"
 
-[[buildpacks]]
-  id = "heroku/ruby"
-  uri = "https://github.com/heroku/heroku-buildpack-ruby/releases/download/cnb-alpha/ruby-buildpack-vCNB-alpha.tgz"
+ [[buildpacks]]
+   id = "heroku/ruby"
+   uri = "https://github.com/heroku/heroku-buildpack-ruby/releases/download/cnb-alpha/ruby-buildpack-vCNB-alpha.tgz"
```
jkutner: why the whitespace here?

jkutner: i think that's probably my fault because they were commented out at one point?

hone:
We need to update the node/java function buildpacks to account for the build plan changes still. We're seeing this error:

```
Selected run image heroku/pack:18
Executing lifecycle version 0.4.0
Using build cache volume pack-cache-e49149178e50.build
Build cache pack-cache-e49149178e50.build cleared
===> DETECTING
[detector] Error: failed to detect: detection failed
ERROR: failed with status code: 6
```

elbandito:
I get the same failure if I comment out all the function build packs in the `functions-builder.toml`
leads me to be believe that somethin outside the buildpacks is causing the failure.

hone:
@elbandito nice job getting the functions stuff working and over the finish line!

PR #25, opened 2019-10-01
Reduce heroku-18 diff

joshwlewis:
This PR attempts to minimize the differences between `pack-18` images and `heroku-18`, with the eventual target to merge a minimal changeset into the `heroku-18` images.


Dockerfile.run
```diff
@@ -1,17 +1,11 @@
...
 ARG pack_gid=1000
-ENV CNB_USER_ID=${pack_uid}
-ENV CNB_GROUP_ID=${pack_gid}
 
-RUN mkdir -p /workspace
-RUN groupadd pack --gid ${pack_gid} && \
-    useradd -u ${pack_uid} -g ${pack_gid} -ms /bin/bash -d /workspace heroku
+RUN groupadd heroku --gid ${pack_gid} && \
+  useradd heroku -u ${pack_uid} -g ${pack_gid} -s /bin/bash -md /workspace && \
+  ln -s /workspace /app
```
jabrown85: I think we can simplify this and remove `/app` from this and the builder. We don't expect to seamlessly support buildpacks and they will break if they don't support symlinks right?

jabrown85: Oh, I see we had it before. I think my question still stands..but I'm not sure. We may be intending to support existing buildpacks more than I imagined.

Dockerfile.build
```diff
@@ -1,21 +1,14 @@
...
 ARG pack_uid=1000
 ARG pack_gid=1000
-ENV CNB_USER_ID=${pack_uid}
-ENV CNB_GROUP_ID=${pack_gid}
 
-RUN mkdir -p /workspace
-RUN groupadd pack --gid ${pack_gid} && \
-    useradd -u ${pack_uid} -g ${pack_gid} -ms /bin/bash heroku
+RUN groupadd heroku --gid ${pack_gid} && \
+  useradd heroku -u ${pack_uid} -g ${pack_gid} -s /bin/bash -md /app
```
jabrown85: I don't think `/app` is where I would make the home directory. I would probably leave it alone.

joshwlewis: That's where $HOME is today during build and inside a running dyno. Doing anything else seems weird.

Dockerfile.run
joshwlewis: I think the original intent was to support legacy buildpacks via the shim. So, sort of seamless maybe? This particular line is only needed for the legacy node buildpack, which seems to hardcode `/app` into $PATH in it's .profile.d.

Dockerfile.build
jabrown85: Nothing happens in the home directory in our current examples. The builder image uses `/home/heroku` but the run image uses `/workspace`. Now we would be using `/app` during build and `/workspace` in the run image right?

FWIW the CF ones use the default home directory of `/home/cnb` in both build and run images.

I wonder what the requirement for a writeable home directory is intended for. @hone?

> The image config's User field is set to a non-root user with a writable home directory.

joshwlewis: I think we could set $HOME to /home/heroku, but we'll still need `/app` to exist and be writeable by the heroku user, at least while we're maintaining support for shimmed buildpacks.

PR #27, opened 2019-10-10
Use Node and NPM cloud native buildpacks for builders

danielleadams:
Replaces the shim buildpack with both the Node.js and NPM cloud native buildpacks built internally. 

Uses the following releases:
https://github.com/heroku/nodejs-engine-buildpack/releases/tag/v0.2.0
https://github.com/heroku/nodejs-npm-buildpack/releases/tag/v0.0.2


jkutner:
@danielleadams can you also remove this line, which prepares the shimmed buildpack? https://github.com/heroku/pack-images/blob/master/buildpacks/install.sh#L43

builder.toml
```diff
@@ -90,5 +94,9 @@ version = "0.4.0"
...
 [[order]]
   [[order.group]]
-    id = "heroku/nodejs"
-    version = "v0.0.5.0.2"
+    id = "heroku/nodejs-engine-buildpack"
+    version = "0.2.0"
+
+  [[order.group]]
+    id = "heroku/nodejs-npm-buildpack"
+    version = "0.0.2"
```
jkutner: i'm surprised that this didn't have/need the heroku/procfile buildpack 🤔

danielleadams: It looks like the test is just checking if the image is successfully build, so I should make sure that it works with the Procfile buildpack too (and add it). I've just been using docker-cli locally for runtime commands

jkutner: ok, i don't think it's absolutely necessary as part of this PR since it wasn't there before.

elbandito:
@danielleadams great work!

Issue #42, opened 2019-11-18
Two /workspace related issues


HansAdema:
Three months ago, in the middle of August, I was able to successfully build an image using Pack and the Heroku images. However, I just tried to build a new image today, but was unable to run the application due to various issues related to the `/workspace` directory.

I'm primarily using the official PHP buildpack, so my error story is based on that. You can verify this with the Heroku getting started example: https://github.com/heroku/php-getting-started

However, I also got errors trying to build the Python getting started image, so I assume all buildpacks are affected by this.

# Issue 1: $HOME is not set correctly

Immediately after building the example image, I am unable to start it.

```bash
$ pack build php-sample
[...]
$ docker run --rm php-sample
vendor/bin/heroku-php-apache2: line 260: php: command not found
This program requires PHP 5.5.11 or newer; check your 'php' command.
```

Checking the PATH inside the container, I see the following:

```
heroku@fb7a0fd3f366:/workspace$ echo $PATH
/home/heroku/.heroku/php/sbin:/home/heroku/.heroku/php/bin:/home/heroku/.heroku/php/sbin:/home/heroku/.heroku/php/bin:/home/heroku/.heroku/php/sbin:/home/heroku/.heroku/php/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/home/heroku/vendor/bin
```

Compare this to the PATH inside the previous, working application:

```
heroku@d384a0aaf899:~$ echo $PATH
/workspace/.heroku/php/sbin:/workspace/.heroku/php/bin:/workspace/.heroku/php/sbin:/workspace/.heroku/php/bin:/workspace/.heroku/php/sbin:/workspace/.heroku/php/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/workspace/vendor/bin
```

The large PATH is largely thanks to the PHP buildpack, which adds some search folders relative to `$HOME`. Which if you compare the old image:

```
heroku@d384a0aaf899:~$ echo $HOME
/workspace
```

Is indeed different from the new image:

```
heroku@fb7a0fd3f366:/workspace$ echo $HOME
/home/heroku
```

This appears to be a consequence of this change: https://github.com/heroku/pack-images/commit/28e9a2cda3d8025db0614a31326599821e0f467d#diff-6f2bd57d9bf5b098ff9f4b368b035e49L12

# Issue 2: The `/app` symlink is missing

[...]

PR #43, opened 2019-11-18
Restore /app symlink and heroku user home directory.

HansAdema:
Fix for issue #42


jkutner:
@HansAdema thanks! we're in a deploy freeze until after Thanksgiving, so we can't release this yet, but we'll get it out as soon as possible.

PR #46, opened 2019-12-12
add salesforce middleware to function builder

jqian2017:
@jkutner , @elbandito , Git repo location of the salesforce middleware build pack is [here](https://github.com/forcedotcom/sf-fx-middleware)

### DIFF
--- a/functions-builder.toml
+++ b/functions-builder.toml
@@ -43,6 +43,7 @@ version = "0.4.0"
   [[order.group]]
     id = "salesforce/sf-fx-middleware"
     version = "0.0.4"
+    optional = true
 
   [[order.group]]
     id = "heroku/node-function"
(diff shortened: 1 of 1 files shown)


functions-builder.toml
```diff
@@ -36,6 +40,10 @@ version = "0.4.0"
     id = "heroku/nodejs-npm-buildpack"
     version = "0.1.2"
 
+  [[order.group]]
+    id = "salesforce/sf-fx-middleware"
+    version = "0.0.4"
```
jkutner: do we want this to be optional like [procfile buildpack](https://github.com/heroku/pack-images/blob/master/builder.toml#L111):

```
optional = true
```

jqian2017: option=true... does it mean if detect fails (return false)... the builder will just build the source w/o the middleware buildpack?

jkutner: yes

elbandito: @jkutner @jqian2017  At first, I was thinking of having an entirely new `[[order]]` for the middleware + npm + nodejs buildpacks, but having `optional = true` seems like a good approach.  I image down the road, if we have other middleware buildpacks that are mutually exclusive from other buildpacks, we'd want to separate those into separate `[[order]]` sections?

PR #47, opened 2019-12-18
Update with the new name of the middleware buildpack and version

jqian2017:


jqian2017:
@jkutner @elbandito , Could one of you add yourself as the reviewer? (since I'm not able to)

We've done some refactoring and renaming of the middleware. so an update to the function builder

jqian2017:
> * superfluous

jqian2017:
@jkutner , thx for the comments...

- `sf` is redundant if the namespace is salesforce
agree
- buildpack is superfluous in general
I tried to follow other buildpack in the function builder to be consistent, they all has buildpack as suffix... if that's the trend.. I can remove it
- `fx` is an internal term, where as fn is a general shorthand for "functions"
I can change to `fn`

@cwallsfdc , what do you think of the name `salesforce/nodejs-fn` (I like that's much shorter)

jqian2017:
oops... I accidentally closed PR (reopen it)

jkutner:
we've [removed `-buildpack` from the node buildpacks](https://github.com/heroku/nodejs-engine-buildpack/blob/master/buildpack.toml#L4) too, but haven't released them yet

cwallsfdc:
I'm open to rules of convention and consistently.

danielleadams:
@jqian2017 just to second what @jkutner said, we decided to drop "buildpack" from the buildpack IDs because its implied as well, but we are keeping repository names with buildpack (ie. `nodejs-engine-buildpack` on GitHub) since it's referencing a code base (and more descriptive of what the code base is). This will keep the `-buildpack` in some of the package urls too.

jqian2017:
we too kept the git repo and the tgz file unchanged, only changed the buildpack ID

jqian2017:
@jkutner @elbandito , could one of you merge for me as I still do not have write access yet.  Thank you

PR #52, opened 2020-01-17
Update function-builder to use decomposed java buildpacks

jkutner:
This matches the `builder.toml`.

@jabrown85 this would mean you don't need to wait for https://github.com/heroku/java-buildpack/pull/9 but then we'd need to apply https://github.com/heroku/java-buildpack/pull/10 to `heroku/jvm` and `heroku/maven`.

### DIFF
--- a/functions-builder.toml
+++ b/functions-builder.toml
@@ -23,8 +23,12 @@ version = "0.4.0"
   uri = "https://github.com/heroku/node-function-buildpack/releases/download/v0.5.0/node-function-buildpack-v0.5.0.tgz"
 
 [[buildpacks]]
-  id = "heroku/java"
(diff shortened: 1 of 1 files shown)


jabrown85:
Makes sense to me! I wondered why they were different when I reviewed the builders yesterday.

jabrown85:
heroku/heroku-buildpack-java#128 and heroku/heroku-buildpack-jvm-common#118 do need to be reviewed and merged in. Any chance we could get those merged in and cut a release before finalizing this?

jabrown85:
I'm less worried about the `provides` bit _right_ now as we are waiting on riff to release before needing that. The stack support is more important I think as it lets us test a great number of things without hacks

jkutner:
I've released the Java and JVM CNBs with the new stack info

PR #54, opened 2020-01-24
Don't create the heroku user with home at /workspace 

jkutner:
…because it polutes the wkdir with bash scripts

### DIFF
--- a/Dockerfile.run
+++ b/Dockerfile.run
@@ -6,7 +6,8 @@ ARG pack_uid=1000
 ARG pack_gid=1000
 
 RUN groupadd heroku --gid ${pack_gid} && \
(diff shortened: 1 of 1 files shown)


Dockerfile.run
```diff
@@ -6,7 +6,8 @@ ARG pack_uid=1000
 ARG pack_gid=1000
 
 RUN groupadd heroku --gid ${pack_gid} && \
-  useradd heroku -u ${pack_uid} -g ${pack_gid} -s /bin/bash -d /workspace -m
+  useradd heroku -u ${pack_uid} -g ${pack_gid} -s /bin/bash -m
 
 LABEL io.buildpacks.stack.id="heroku-18"
 USER heroku
+ENV HOME /app
```
jkutner: I don't actually know if this will suffice. The problem will be the PHP buildpack.

PR #57, opened 2020-01-28
add yarn buildpacks and groups

danielleadams:
Adds the yarn buildpack to both the functions-builder and builder files.

For the groups, the builders will detect for first yarn group, and then fall back on using npm.


hone:
@danielleadams just curious why you detect `yarn` first?

danielleadams:
@hone npm is the default (since it doesn't require an additional install), so the yarn.lock file should be checked first to determine whether the yarn group should be used; if yarn.lock is not present, builder uses npm buildpack.

PR #65, opened 2020-02-14
Lifecycle 0.6.1

jkutner:
Also fixes a mistake in the Gradle group where the wrong `heroku/procfile` version is used

### DIFF
--- a/builder.toml
+++ b/builder.toml
@@ -4,7 +4,7 @@ build-image = "heroku/pack:18-build"
 run-image = "heroku/pack:18"
 
 [lifecycle]
-version = "0.5.0"
+version = "0.6.0"
 
 [[buildpacks]]
   id = "heroku/maven"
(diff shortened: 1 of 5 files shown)


builder.toml
```diff
@@ -4,7 +4,7 @@ build-image = "heroku/pack:18-build"
 run-image = "heroku/pack:18"
 
 [lifecycle]
-version = "0.5.0"
+version = "0.6.1"
```
jabrown85: Will bumping this break any buildpacks that exist outside of our builders that we know of?

jkutner: not that i know of. but it could. probably more reason to version the image beyond `:18`

qw3r: I like the idea of having versions beyond `:18`, it would make us simply fallback to previous "version" of the package, which we are not able to do now :(

jkutner:
I keep running into problems with the buildpacks, and I think we're going to need to update the cnb-shim so that it runs the v2 buildpacks in a tmp (writeable) dir

jkutner:
This update to cnb-shim introduces https://github.com/heroku/cnb-shim/commit/eece77504e9fd3fdf78b029c0bd804373c53d02d

PR #70, opened 2020-02-20
Remove unused go code

jabrown85:
This was used in the past when we built the lifecycle binaries

[W-7246185](https://gus.lightning.force.com/a07B0000007Hr43IAC)


jabrown85:
@joshwlewis I was going to add snyk but realized we aren't actually using/building the go code anymore. I'm not sure how much value snyk has for bash/dockerfile based repos.

PR #78, opened 2020-03-18
upgrade to lifecycle 0.7.2. Some perf improvements to launcher.

hone:
[Lifecycle 0.7.2 Release Notes](https://github.com/buildpacks/lifecycle/releases/tag/v0.7.2)

Signed-off-by: Terence Lee <hone02@gmail.com>


hone:
Looks like we need to update `pack-runner` to use a later version of `pack`.

jkutner:
@hone yea that happened the last time we upgraded the version. Maybe we should build it on the fly

hone:
I updated the `pack-runner` image to `0.9.0`. In another PR, we should maybe move to the Circle Orb

hone:
the Ruby, Python, and PHP buildpacks depend on `STACK` to be set in the environment. lifecycle 0.7 breaks this because it brings it in according to [the spec](https://github.com/buildpacks/spec/blob/master/buildpack.md#provided-by-the-platform). This means that env vars that are set in the container won't just get propagated through. We either need to update the buildpacks to set `clear-env` to `false` or figure something else out here.

hone:
Though maybe it's better to just update cnb-shim to set `STACK=CNB_STACK_ID`? I'm not sure why the ruby buildpack failed since we're [setting `STACK`](https://github.com/heroku/heroku-buildpack-ruby/blob/master/bin/build#L11-L12).

go.mod
```diff
@@ -0,0 +1,11 @@
...
+
+require (
+	github.com/BurntSushi/toml v0.3.1
+	github.com/buildpack/lifecycle v0.4.1-0.20191016200822-f0a279f287ef
+	github.com/buildpack/pack v0.5.0 // indirect
+	github.com/buildpacks/pack v0.9.0 // indirect
+	github.com/gotestyourself/gotestyourself v2.1.0+incompatible // indirect
+)
+
+go 1.13
```
jkutner, line 11: did you mean to add this back?

PR #82, opened 2020-05-08
Add meta-buildpacks and replace shim

jkutner:
This change does two things:

* Removes the cnb-shim logic, and uses the `cnb-shim.herokuapp.com` service instead
* Adds meta-buildpacks for `heroku/nodejs` and `heroku/java`

The meta-buildpacks solve the problem where a user runs into:

```
$ pack build myimage --buildpack heroku/nodejs
...
Status: Downloaded newer image for heroku/nodejs:latest
ERROR: extracting buildpacks from heroku/nodejs: could not find label io.buildpacks.buildpackage.metadata
```

This happens because there is no `heroku/nodejs` buildpack in the builder image, which causes Pack to try and download the `docker.io/heroku/nodejs` image, which isn't a buildpack.

## Future Work

The `heroku/nodejs` and `heroku/java` buildpacks should be packaged into a buildpackage such that they include their own dependencies (and we don't need to put their dependencies in the `builder.toml`). We will wait on this until there is a Buildpack Registry where we can store/pull them.


jkutner:
@danielleadams in the function builder, do you care if the node buildpack has the id `heroku/nodejs`? If not, then we can simply give the no-yarn-buildpack a different id (or continue to only use the order/groups. If you'd prefer that it have the same id, then that's a pickle.

I'd recommend giving them different IDs since they have different behavior, and users of the function builder are less likely to customize their buildpacks.

danielleadams:
@jkutner okay, makes sense. It's not a permanent need, but during development there may be a need for 2 different Node buildpacks.

PR #85, opened 2020-05-19
Upgrade lifecycle to 0.7.4 (builder.toml only)

jkutner:
Also, remove redundant versions defined in `builder.toml`


hone:
TIL that `version` is optional.

Is it a problem that the functions builder and normal builder use different lifecycles?

PR #89, opened 2020-05-20
replace forked heroku/node-function buildpack with riff

elbandito:
**NOTE: This PR has a dependency on https://github.com/forcedotcom/nodejs-sf-fx-buildpack/releases/tag/v1.5.0

This PR is responsible for removing our dependency on our forked `heroku/node-function` buildpack/invoker, and instead will allow us to use the upstream version of riff.  

Additionally, short-term changes were made to satisfy the `RIFF=1` requirement in order for the `detect` phase to pass while building a function.  Having the ENV var stored in `/plaform` allows us to truly hide it from the user, and will make the removal process seamless.

### DIFF
--- a/Dockerfile.riff.build
+++ b/Dockerfile.riff.build
@@ -1,5 +1,0 @@
-FROM heroku/pack:18-build
-
-USER root
-RUN mkdir -p /platform/env && echo "1" > /platform/env/RIFF
-USER heroku
--- a/Dockerfile.riff.build
+++ b/Dockerfile.riff.build
@@ -2,3 +2,4 @@ FROM heroku/pack:18-build
 
 USER root
 RUN mkdir -p /platform/env && echo "1" > /platform/env/RIFF
+USER heroku
--- a/Dockerfile.riff.build
+++ b/Dockerfile.riff.build
@@ -1,0 +1,4 @@
+FROM heroku/pack:18-build
+
+USER root
+RUN mkdir -p /platform/env && echo "1" > /platform/env/RIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ SHELL=/bin/bash -o pipefail
 build:
 	@docker pull heroku/heroku:18-build
 	@docker build -f Dockerfile.build -t heroku/pack:18-build .
+	@docker build -f Dockerfile.riff.build -t heroku/pack:18-riff-build .
 	@docker build -f Dockerfile.run -t heroku/pack:18 .
 	@pack create-builder heroku/buildpacks:18 --builder-config builder.toml --no-pull
 	@pack create-builder heroku/functions-buildpacks:18 --builder-config functions-builder.toml --no-pull
@@ -12,6 +13,7 @@ build:
 	@docker tag heroku/functions-buildpacks:18 heroku/functions-buildpacks:latest
 
 publish: build
+	@docker push heroku/pack:18-riff-build
 	@docker push heroku/pack:18-build
 	@docker push heroku/pack:18
 	@docker push heroku/buildpacks:18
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -5,13 +5,15 @@ SHELL=/bin/bash -o pipefail
...
 	@docker build -f Dockerfile.build -t heroku/pack:18-build .
+	@docker build -f Dockerfile.riff.build -t heroku/pack:18-riff-build .
 	@docker build -f Dockerfile.run -t heroku/pack:18 .
 	@pack create-builder heroku/buildpacks:18 --builder-config builder.toml --no-pull
 	@pack create-builder heroku/functions-buildpacks:18 --builder-config functions-builder.toml --no-pull
 	@docker tag heroku/buildpacks:18 heroku/buildpacks:latest
 	@docker tag heroku/functions-buildpacks:18 heroku/functions-buildpacks:latest
 
 publish: build
+	@docker push heroku/pack:18-riff-build
```
joshwlewis, line 16: What's the reasoning for adding another image? This seems like a lightweight change that would make sense in `pack:18-build` and probably be fine for `pack:18` for that matter.

elbandito: @joshwlewis I'm being paranoid, maybe too much.  I could be swayed to add this to the exiting `pack:18-build` we if don't think it would cause any issues having `RIFF=1` in all of the images.  It should only be for a short amount of time.

joshwlewis: Why only a short amount of time? Are there upstream changes coming?

jkutner: I'd rather not have it in the `heroku/pack:18-build` image. There could be unforeseen consequences to people using `heroku/buildpacks:18`. Image layers are cheap (especially this one) so I don't mind the extra tag if need be. I haven't thought it out well though.

elbandito: @joshwlewis yeah, Ben Hale is planning something that should allow this to go away.  Once that mechanism is in place we should be able to remove this.

hone: This just sets `RIFF=1` for every image. I don't think this would cause issues for most people, but probably better to be safe atm than sorry.

Dockerfile.riff.build
```diff
@@ -0,0 +1,4 @@
+FROM heroku/pack:18-build
+
+USER root
```
jabrown85, line 3: Do we need another `USER heroku` beneath the next `RUN`? Does this image start the same as `18-build`? The last `USER` directive is important sometimes, so I just want to make sure we aren't going to accidentally be running `exec` commands as `root`.

`docker run -it heroku/pack:18-riff-build whoami`

functions-builder.toml
```diff
@@ -45,12 +49,12 @@ version = "0.6.1"
     version = "0.2.0"
 
   [[order.group]]
-    id = "salesforce/nodejs-fn"
-    version = "1.4.3"
+    id = "io.projectriff.node"
+    version = "0.3.1-BUILD-SNAPSHOT"
 
   [[order.group]]
-    id = "heroku/node-function"
```
jabrown85, line 52: 🎉

Makefile
jabrown85: Yeah, I think I lean toward the extra image for safety + it will ensure we keep poking the riff team to get it moved to `requires` for consumers like us. 

Just thinking out loud...root buildpacks wouldn't allow us to fix this either right? Since the env var needs to be done prior to `detect`.

hone:
@jabrown85 yeah, `detect` should still be run in parallel so that wouldn't help, but I wonder if that's something we need to think about.

Makefile
elbandito: Yeah, this is required prior to `detect`.

PR #93, opened 2020-06-02
create meta buildpack for evergreen node function

elbandito:
This is a metabuildpack for the evergreen function.

https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008PUfQIAW/view


jkutner:
Is `evergreen/node-function` the buildpack ID we want people to type? I think we might be able to shorten/simplify it. Options:

* `evergreen/node` (and deal with whatever non-functions look like later, possibly letting the buildpack itself figure it out with optional bps)
* `evergreen/node-fn`

elbandito:
> Is `evergreen/node-function` the buildpack ID we want people to type? I think we might be able to shorten/simplify it. Options:
> 
> * `evergreen/node` (and deal with whatever non-functions look like later, possibly letting the buildpack itself figure it out with optional bps)
> * `evergreen/node-fn`
> * `evergreen/js` (do we want to stick with "node" even if they aren't really using it?)
> 
> /cc @danielleadams what do you think about "node" versus "js"?

I just threw something up for the time being, so I'm completely open to other opinions on naming.

jabrown85:
Question about meta buildpacks...can we have multiple detection groups? If so maybe we want `evergreen/function` and that would work for `node` and `java`+ right?

elbandito:
> Question about meta buildpacks...can we have multiple detection groups? If so maybe we want `evergreen/function` and that would work for `node` and `java`+ right?

@jabrown85 are you staying just combine both the original Java and Node groups into a `evergreen/function` metabuildpack?

jabrown85:
> > Question about meta buildpacks...can we have multiple detection groups? If so maybe we want `evergreen/function` and that would work for `node` and `java`+ right?
> 
> @jabrown85 are you saying just combine both the original Java and Node groups into a `evergreen/function` metabuildpack?

Maybe! If we can have multiple ordered groups in a single meta buildpack. That would mean users would execute `--buildpack evergreen/function --buildpack mycustom/buildpack` regardless of the language.

I guess I'm wondering what value there is to having language specific buildpack identifiers at this stage. What is the use case for needing to explicitly have `node` or `java`? I guess a customer repo having a `package.json` but you are really doing `java` functions?

We could also consider multiple meta buildpacks to allow specificity once we have a second language.

`evergreen/fn` contains `evergreen/fn:js` && `evergreen/fn:java`

jkutner:
+1 to `evergreen/fn` containing other metabuildpacks for node and java

danielleadams:
`evergreen/fn` sounds good to me because then that's assuming that there's optional TypeScript buildpack added too.

elbandito:
@jkutner how do you feel about just removing the `java` stuff for now,  since it doesn't work anyway?  Then, we can just have the single metabuildpack for evergreen node functions called `evergreen/fn`, which we can later split it into sub-metabuildpacks once we start adding additional language support (I think Jesse may have alluded to this as well).

jkutner:
@elbandito can we keep the java buildpacks in the builder but remove them from the order (just in case we're missing somewhere that they're used)?

buildpacks/evergreen_function/buildpack.toml
```diff
@@ -0,0 +1,23 @@
...
+    id = "heroku/nodejs-engine"
+    version = "0.4.3"
+
+  [[order.group]]
+    id = "heroku/nodejs-npm"
+    version = "0.2.0"
+
+  [[order.group]]
+    id = "io.projectriff.node"
+    version = "0.3.1-BUILD-SNAPSHOT"
```
jabrown85, line 19: Is there a non-snapshot release of this?

elbandito: @jabrown85 I believe there is now with the changes that Ben mentioned.  We just need to re-visit how Ben's changes will affect the other buildpacks with the renaming of `[provides]` and `[requires]` stuff, and we'll need to also the `http-streaming-adapter` buildpack to the list.   I did add a spike in the backlog (not in this Sprint, but can still prioritize for this Sprint if we want).

https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B000000893wxIAA/view

PR #101, opened 2020-06-29
add typescript buildpack to heroku/nodejs builder

danielleadams:
Adds TypeScript buildpack from release v0.0.2. For now, the buildpack is only available in the Heroku meta buildpack.

This pull request also includes an added test in CI. The getting started guide (https://github.com/danielleadams/typescript-getting-started) is being reviewed by Salesforce OSS, so it will be able to use the same Job once the repo gets approved.

## Context
As we move towards creating a great Node experience for JavaScript developers, we want to make sure that the TypeScript experience is just as good. By adding supporting a TS buildpack, we can make data and feedback-based enhancements to the DX for developers using CNBs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -91,6 +91,22 @@ jobs:
       - run: docker load < pack-18-run.tar
       - run: docker load < buildpacks-18.tar
       - run: pack build pack-getting-started --path getting_started --builder heroku/buildpacks:18 --no-pull
+  test-typescript-getting-started-guide:
+    docker:
+      - image: heroku/pack-runner:latest
+    steps:
+      - run: git clone https://github.com/danielleadams/typescript-getting-started.git getting_started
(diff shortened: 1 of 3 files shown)


hone:
@danielleadams do we want this as part of the `functions-builder.toml` too?

danielleadams:
@hone I figured not for right now assuming that version of the buildpack is being used for Evergreen Dev Preview. I could though, and it wouldn't change the behavior of anyone already using TypeScript.

hone:
yeah, it's fine for them to diverge, but I worry about having them too divergent for long periods if we intend to roll them together plus TypeScript is supported in Salesforce Functions.

danielleadams:
@hone yeah, that makes sense. I have to come back and change the test anyway once the typescript repo gets approved, so I can add it when I change the test.

PR #118, opened 2020-09-08
chore: upgrade lifecycle to v0.9.1

jabrown85:
https://github.com/buildpacks/lifecycle/releases/tag/v0.9.1

### DIFF
--- a/builder.toml
+++ b/builder.toml
@@ -4,7 +4,7 @@ build-image = "heroku/pack:18-build"
 run-image = "heroku/pack:18"
 
 [lifecycle]
-version = "0.8.0"
+version = "0.9.1"
 
 [[buildpacks]]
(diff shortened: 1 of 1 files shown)


builder.toml
```diff
@@ -4,7 +4,7 @@ build-image = "heroku/pack:18-build"
 run-image = "heroku/pack:18"
 
 [lifecycle]
-version = "0.8.0"
```
joshwlewis, line 7: Should we do this in `functions-builder.toml` too?

jabrown85: I thought about it..but then remembered you had an open item to merge the two builders into one. So I was thinking of ignoring the other builder for now.

joshwlewis (approved):

PR #119, opened 2020-09-08
Use newer canaries in CI

joshwlewis:
We previously leaned on https://github.com/heroku/evergreen-hilltop-theatre-terraform as canaries for validating function builders. That repository is not seeing updates as event invocation is scheduled for later delivery. We should use canaries that are actively being updated and tested instead.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -231,15 +231,15 @@ workflows:
       - test-evergreen-canary:
           name: test-canary-https-invocation-hello
           url: canary-https-invocation
-          path: hello
-          fingerprint: "d4:04:b8:b2:2d:24:f3:6e:b2:5d:5b:da:c1:82:65:00"
+          path: hello-function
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -229,30 +229,16 @@ workflows:
           requires:
             - create-service-builder
       - test-evergreen-canary:
-          name: test-hilltop-sms
-          url: hilltop-theatre-terraform
-          path: sms-ticket-function
+          name: test-canary-https-invocation-hello
+          url: canary-https-invocation
+          path: hello
```
jabrown85, line 234:
```suggestion
          path: src/hello-function
```
Maybe?

joshwlewis: "src/" is implied in the job, but I did miss the "-function" part!

jabrown85 (approved):

PR #126, opened 2020-09-18
Stop publishing functions-buildpacks

joshwlewis:
Now that heroku/buildpacks:18 can build functions, we don't need `functions-buildpacks` anymore. This PR stops us from automatically building it on commits or nightly. Note that the image is still on [dockerhub](https://hub.docker.com/r/heroku/functions-buildpacks) (and listed as deprecated), but will no longer be updated.

[GUS Item](https://gus.lightning.force.com/a07B0000008UsrfIAC)


jabrown85 (approved):

elbandito (approved):
:surfer:

PR #130, opened 2020-10-13
chore: Bump nodejs-sf-fx-buildpack to v2.0.1

joshwlewis:
This brings in the changes from forcedotcom/nodejs-sf-fx-buildpack/pull/63.


jabrown85 (approved):
This looks good! The CLI was updated, so the new functions generated with the new sdk should be good to go.

jabrown85 (approved):

PR #131, opened 2020-10-14
update nodejs to 0.5.0 and npm to 0.3.0

danielleadams:
Upgrades Node.js buildpacks.

Follow up:
- upgrade nodejs buildpack to 0.6.0

### DIFF
--- a/builder.toml
+++ b/builder.toml
@@ -48,11 +48,11 @@ version = "0.9.1"
 
 [[buildpacks]]
   id = "heroku/nodejs-engine"
-  uri = "https://github.com/heroku/nodejs-engine-buildpack/releases/download/v0.4.4/nodejs-engine-buildpack-v0.4.4.tgz"
+  uri = "https://github.com/heroku/nodejs-engine-buildpack/releases/download/v0.5.0/nodejs-engine-buildpack-v0.5.0.tgz"
 
 [[buildpacks]]
   id = "heroku/nodejs-npm"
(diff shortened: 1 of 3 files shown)


jabrown85:
@danielleadams Any risky bits in the downstream buildpacks to be worried about?

jabrown85 (approved):

danielleadams:
@jabrown85 There shouldn't be if devDependencies vs. dependencies in the `package.json` are being used to separation build/runtime dependencies. Here's the changelog for the NPM buildpack:

https://github.com/heroku/nodejs-npm-buildpack/blob/main/CHANGELOG.md#030-2020-09-16

It will remove devDependencies from the runtime environment if NODE_ENV is set to `production`, so I am worried about breakage downstream only in the event that a user's runtime dependencies were added to their devDependencies on accident AND they have set their NODE_ENV to production.

I have a follow up PR that will set NODE_ENV to production in the build environment unless otherwise specified ([`nodejs-engine/0.6.0`](https://github.com/heroku/nodejs-engine-buildpack/blob/main/CHANGELOG.md#060-2020-10-13)), that will only cause issues in the event that runtime dependencies ended up in devDependencies.

danielleadams:
(I'll do a better job of adding all this context in the descriptions moving forward too.. this would be helpful to include for reference.)

jabrown85:
Sounds reasonable to me. Thanks! Feel free to merge. Once CI builds the new buildpack will be live for all users who pull a new builder.

PR #133, opened 2020-10-27
Start building heroku/pack:20 and heroku/pack:20-build

edmorley:
The first step for Heroku-20 support is to start building and publishing the `heroku/pack:20` and `heroku/pack:20-build` stack images.

Once these are available, CNB authors will be able to test their buildpacks against the new stack, and in a later PR we'll start publishing the `heroku/buildpacks:20` builder with the updated CNBs.

For more on the remaining Heroku-20 tasks, see:
https://salesforce.quip.com/LbhlAFPd1rvR#fCAACA6FA2X

Closes [W-8270099](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008hLTWIA2/view).

### DIFF
--- a/Dockerfile.run
+++ b/Dockerfile.run
@@ -1,4 +1,5 @@
-FROM heroku/heroku:18
+ARG BASE_IMAGE
+FROM ${BASE_IMAGE}
 
 RUN ln -s /workspace /app
 
@@ -8,6 +9,7 @@ ARG pack_gid=1000
 RUN groupadd heroku --gid ${pack_gid} && \
   useradd heroku -u ${pack_uid} -g ${pack_gid} -s /bin/bash -m
 
-LABEL io.buildpacks.stack.id="heroku-18"
+ARG STACK
+LABEL io.buildpacks.stack.id="${STACK}"
 USER heroku
 ENV HOME /app
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,16 +5,18 @@ jobs:
       dockerfile:
         description: "Dockerfile to build"
         type: string
+      base_image:
+        description: "The Docker image to be used in the FROM line of the Dockerfile"
+        type: string
+      stack_name:
+        description: "The name of the stack"
+        type: string
       image_file:
         description: "Local image archive file name"
         type: string
       image_tag:
         description: "Remote image tag name"
         type: string
-      target:
-        description: "Docker build target"
-        type: string
-        default: ""
     docker:
       - image: heroku/pack-runner:latest
     steps:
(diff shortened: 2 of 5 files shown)


Dockerfile.run
```diff
@@ -8,6 +9,7 @@ ARG pack_gid=1000
 RUN groupadd heroku --gid ${pack_gid} && \
   useradd heroku -u ${pack_uid} -g ${pack_gid} -s /bin/bash -m
 
-LABEL io.buildpacks.stack.id="heroku-18"
+ARG STACK
+LABEL io.buildpacks.stack.id="${STACK}"
```
jkutner, line 13: I don't want to block @edmorley on this, but switching to `heroku/heroku:20` might be a good opportunity for pack-images and our CNBs to move away from `stack = "heroku-20"` and towards `stack = "io.buildpacks.stacks.bionic"` plus mixins.

jkutner: Example:
https://github.com/paketo-buildpacks/stacks/blob/main/bionic/cnb/run/Dockerfile#L13-L14

edmorley: @jkutner Should I ask the CNB maintainers to hold off adding `stack = "heroku-20"` then? I've just filed a bunch of work items asking them to add it, and https://github.com/heroku/procfile-cnb/pull/5 has already merged.

Or is such a move something we can do in a non-breaking way for both heroku-18 and heroku-20 in the future? Since `builder.toml` contains versioned CNB URLs, it seems we can switch over the builder to use the new IDs in one go in the future, without breakage?

jkutner: We should have the buildpacks add `heroku-20`, so those work items are 👌 We can migrate without breaking anything.

(Sidenote: Actually, we really want to move `io.buildpacks.stacks.focal`)

jabrown85: If the referenced PRs aren't merged, we might consider adding `io.buildpacks.stacks.focal` and `io.buildpacks.stacks.bionic` as supported stacks in those buildpack PRs. It may save some time in the future and unless someone is building their own builder, it shouldn't have any consequences.

jkutner: @jabrown85 but that would be a lie in most cases, because they also require some mixins. maybe better to wait until the buildpack owners can add both?

jabrown85: ah, yeah. That is the part that was _super_ annoying when we tried to first make the switch to bionic. It's whack-a-mole trying to figure out which mixins you require to be present.

edmorley: Yeah I think since we don't yet have the list of mixins, it would be best to hold off on buildpacks declaring support for `io.buildpacks.stacks.focal`.

Since the conversation about changing stack name/label has moved to email, and that we don't want to block on it, are we good to call this PR thread resolved, and to continue with this PR? :-)

jkutner: Yes, we should move forward with this PR

jabrown85 (approved):

edmorley:
I'm getting a:

> Merging is blocked
> The base branch requires all commits to be signed

I don't have GPG signing set up locally, but since I was going to use "squash and merge", GitHub automatically signs the commit that ends up on master, per:
https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/about-commit-signature-verification#about-commit-signature-verification

As such the GitHub check seems broken?

jabrown85:
> As such the GitHub check seems broken?

I noticed the same. We were told to turn that check on by security. I think you'll need to amend your commits with a gpg signature. Sorry :(

edmorley:
No worries - good excuse to enable GPG signing locally too :-)

edmorley:
Images published:
https://app.circleci.com/pipelines/github/heroku/pack-images/737/workflows/c627ad42-0928-4489-834b-f3d6e91d13ac
https://hub.docker.com/r/heroku/pack/tags/?page=1&name=20

PR #137, opened 2020-11-16
Update Pack to 0.15.0, lifecycle to 0.9.3

joshwlewis:
This PR updates pack to 0.15.0 and lifecycle to 0.9.3.

It also drops the custom CI image in favor of inline pack installation, since that workflow was manual and completely undocumented.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,7 +65,7 @@ jobs:
       - run: docker pull gcr.io/projectriff/node-function:0.6.1
       - run: docker load < $(ls /tmp/workspace/pack-*-build.tar | head -1)
       - run: docker load < $(ls /tmp/workspace/pack-*-run.tar | head -1)
-      - run: pack create-builder << parameters.image_tag >> --builder-config << parameters.builder_toml >> --no-pull
+      - run: pack create-builder << parameters.image_tag >> --config << parameters.builder_toml >> --no-pull
       - run: docker save << parameters.image_tag >> > /tmp/workspace/<< parameters.image_file >>
       - persist_to_workspace:
           root: /tmp/workspace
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -42,17 +42,30 @@ jobs:
         description: "Remote builder image name"
         type: string
     docker:
-      - image: heroku/pack-runner:latest
+      - image: circleci/golang:1.15
```
jabrown85, line 45: There are a couple ways we could explore to make this even better.

1. Use the `buildpacksio/pack:0.15.0` image. I see a lot of docker load though, so that might not work out. I'm not sure what all is in that image. It might _just_ be `pack` itself. We could still turn all the `pack <x>` commands into `docker run -v ... buildpacksio/pack <x>`.
2. Use the `apt-get install pack-cli`

jabrown85 (approved):

jabrown85 (changes requested):
Need a makefile update, actually. Remove `build-ci`

.circleci/config.yml
joshwlewis: `buildpacks/pack` uses `index.docker.io/paketobuildpacks/run:tiny-cnb` as a run image, which is a super minimal distroless bionic image. So yeah, it's just `pack` in there really. I'll give docker run a shot here.

joshwlewis: `docker run buildpacksio/pack` was getting hairy, because we need access to the remote docker daemon from inside the pack container. Mounting /var/run/docker.sock was giving permission errors. I think we'd need to pass down `DOCKER_HOST` and the other host vars along with it, and I'm not sure it's worth the complexity.

`sudo add-apt-repository ppa:cncf-buildpacks/pack-cli` seems to work, but it's slow since we have to run `apt-get update` which fetches more repositories than we need.

PR #140, opened 2020-12-07
Update heroku/jvm to v113

Malax:
Updates heroku/jvm to v113. Note that this version changed the CNB version from `0.1` to `0.1.0`.


Malax:
Closing in favor of #148

PR #145, opened 2021-01-04
Bump lifecycle version to 0.10.1

jabrown85:
- Update to lifecycle 0.10.1 on all our builders
- Update to pack 0.15.1 in circle ci and update cli usage
- Update Makefile to successfully build with updated version of pack while only pulling images not on the local machine
- Updated builder image references to use docker:// as suggested by pack upon failure

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,9 +53,9 @@ jobs:
       - run: &install-pack
           name: "Install pack"
           command: |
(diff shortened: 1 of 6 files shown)


jabrown85:
I tested building a Salesforce function with pack at 0.15.1 and with the sfdx CLI and they both produced running images.

joshwlewis (approved):

PR #146, opened 2021-01-04
Ensure getting-started guide containers start successfully

jabrown85:
We've had issues where buildpack upgrades create containers successfully, but then the container immediately crashes. A small smoke test after each getting-started guide should ensure the images at least remain running and prevent some bad releases.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,6 +93,9 @@ jobs:
       - run: docker load < /tmp/workspace/buildpacks-<< parameters.stack-tag >>.tar
       - run: pack trust-builder heroku/buildpacks:<< parameters.stack-tag >>
       - run: pack build pack-getting-started --path getting_started --builder heroku/buildpacks:<< parameters.stack-tag >> --pull-policy never
+      - run: docker run --name getting-started -d -e PORT=8080 pack-getting-started
+      - run: sleep 5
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -93,6 +93,9 @@ jobs:
       - run: docker load < /tmp/workspace/buildpacks-<< parameters.stack-tag >>.tar
       - run: pack trust-builder heroku/buildpacks:<< parameters.stack-tag >>
       - run: pack build pack-getting-started --path getting_started --builder heroku/buildpacks:<< parameters.stack-tag >> --pull-policy never
+      - run: docker run --name getting-started -d -e PORT=8080 pack-getting-started
+      - run: sleep 5
+      - run: docker exec getting-started true 2>/dev/null || (echo not running && docker logs getting-started && exit 1)
```
joshwlewis, line 98: I think this would have caught the boot failures from before break, so I'm +1. It doesn't totally prove that the web server is running and serving requests on the designated port. We could do that with `curl` for another level of assurance.

joshwlewis (approved):

.circleci/config.yml
jabrown85: yeah, I didn't want this repo being the end-all for testing each permutation. Each time we update the output of any default project template we'd have to change this repo in some way. This at least ensures the buildpack groups are setup correctly enough to have a container that starts something by default and continues running for some amount of time.

PR #148, opened 2021-01-22
Update JVM buildpack URIs, Remove heroku/java buildpack

Malax:
JVM CNB buildpacks are now hosted on Amazon's ECR and are available via the CNB registry. Since `pack` does not support `urn:cnb:registry:*` URIs yet, direct `docker://` URIs are used for now.

JVM buildpacks that had been shimmed on-the-fly are now shimmed once per V2 release. This also gives us more useful version numbers for those.

The `heroku/java` buildpack was moved to https://github.com/heroku/buildpacks-jvm/tree/main/meta-buildpacks/java where it is now maintained along other JVM CNBs.

Since `heroku/java` uses `heroku/procfile-cnb`@`0.6.1`, we had a version conflict in the builders. I took the liberty to update `heroku/procfile-cnb` to `0.6.1` for everything else as well. Alternatively, we could've annotated all `heroku/procfile` uses with an explicit version.

In CI, building the service image failed since `pack create-builder` was invoked with `--pull-policy never`. I could've added a `docker pull` step before the create-builder step for each buildpack. This was previously done before for `streaming-http-adapter:0.1.3` and `node-function:0.6.1`. 

However, I think it's too much hassle to update the buildpack docker URIs in the CI config as well as the `builder.toml` every time we have a buildpack update. So I changed the pull policy to `if-not-present`. Happy to discuss this further. (@jabrown85, you added the `docker pull`s - you probably have an opinion on this I assume) The actual Makefiles also use `if-not-present`.


builder-18.toml
```diff
@@ -8,36 +8,36 @@ version = "0.10.1"
 
 [[buildpacks]]
   id = "heroku/maven"
-  uri = "https://github.com/heroku/heroku-buildpack-java/releases/download/v68/heroku-java-cnb-v68.tgz"
+  uri = "docker://public.ecr.aws/r2f9u0w4/heroku-maven-buildpack:0.2.0"
```
jabrown85, line 11: What's the `r2f9u0w4` bit?

.circleci/config.yml
```diff
@@ -61,11 +61,9 @@ jobs:
       - attach_workspace:
           at: /tmp/workspace
       - setup_remote_docker
-      - run: docker pull gcr.io/projectriff/streaming-http-adapter:0.1.3
-      - run: docker pull gcr.io/projectriff/node-function:0.6.1
       - run: docker load < $(ls /tmp/workspace/pack-*-build.tar | head -1)
       - run: docker load < $(ls /tmp/workspace/pack-*-run.tar | head -1)
-      - run: pack create-builder << parameters.image_tag >> --config << parameters.builder_toml >> --pull-policy never
+      - run: pack create-builder << parameters.image_tag >> --config << parameters.builder_toml >> --pull-policy if-not-present
```
jabrown85, line 66: I _think_ I'm ok with this now. Originally `pack` only support `never` via `--no-pull`. We needed to not `pull`, otherwise the act of creating a builder would use the published build/run images and not use the ones built in the CI run.

I wish `pack` had a pull policy for buildpacks vs image pull policies. If we make a mistake and `docker load` gets removed or mistyped we could be publishing builders on old images.

builder-18.toml
```diff
@@ -8,36 +8,36 @@ version = "0.10.1"
...
+  uri = "docker://public.ecr.aws/r2f9u0w4/heroku-jvm-buildpack:0.1.2"
 
 [[buildpacks]]
   id = "heroku/java"
-  uri = "buildpacks/heroku_java"
+  uri = "docker://public.ecr.aws/r2f9u0w4/heroku-java-buildpack:0.1.3"
+
+[[buildpacks]]
+  id = "heroku/gradle"
+  uri = "docker://public.ecr.aws/r2f9u0w4/heroku-gradle-buildpack:0.0.34"
```
jabrown85, line 23: Do you have a versioning plan if you need to cut a new release of the same base version but with an updated shim? We were discussing moving all the shimmed buildpacks to a registry like this. Will you just overwrite `0.0.34` or bump the version in some way?

jabrown85 (commented):
I asked a few questions, but this looks great!

builder-18.toml
Malax: That's our ECR repository identifier. I've applied for a custom one `heroku-buildpacks`, but their manual approval process takes longer than expected. The old identifier shown here stays valid indefinitely.

.circleci/config.yml
Malax: Thanks for the explanation behind `never`, I was wondering why we'd do it this way. Makes sense. If we want to avoid this potential issue, we can do smth. like this before calling `create-builder`:

```bash
#!/usr/bin/env bash

for buildpack_uri in $(yj -t <builder-18.toml | jq -r '.buildpacks[].uri | select(startswith("docker://"))'); do
    docker pull "${buildpack_uri}"
done
```

With this, we can keep `never` but don't have to maintain two lists of images to pull. What do you think @jabrown85?

builder-18.toml
Malax: Good point. There is no versioning plan for this. I think we would just overwrite should that case ever arise. The chance of this happening is quite slim I think. And if it were to become a problem, we can always change the versioning scheme it so smth. like `1.${V2_VERSION}.${SHIM_VERSION}` without violating semver.

Also, at least for the JVM buildpacks, I'm trying to move away from shimmed V2 buildpacks ASAP. Happy to change the versioning scheme if you believe it to be a problem though.

.circleci/config.yml
jabrown85: I think I'm ok leaving it as you have it right now. I like the little bash helper, but it won't help buildpacks.io registry published images.

jabrown85 (approved):

jkutner:
`pack` should support the `urn:cnb:registry` URIs (maybe it needs a version bump?).

Regardless, it may be preferable to use the `docker://` URIs until we can get a proper security review of the Registry.

Malax:
I retroactively signed my commits and fixed a (new) conflict, no other changes have been made between approval and merging this PR.

PR #150, opened 2021-02-01
Revert "Update JVM buildpack URIs, Remove heroku/java buildpack"

jabrown85:
Reverts heroku/pack-images#148

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,9 +61,11 @@ jobs:
       - attach_workspace:
           at: /tmp/workspace
       - setup_remote_docker
+      - run: docker pull gcr.io/projectriff/streaming-http-adapter:1.4.0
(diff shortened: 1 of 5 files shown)


Malax (approved):

schneems:
Is this revert related to https://discuss.hashicorp.com/t/error-multiple-versions-of-heroku-procfile-must-specify-an-explicit-version/20390 ?

jabrown85:
> Is this revert related to https://discuss.hashicorp.com/t/error-multiple-versions-of-heroku-procfile-must-specify-an-explicit-version/20390 ?

Yes. We found that same error internally. The bundled version of `pack` does not deal with the builder versions in the same way, this was found in our `sfdx` tooling but is also present in the `hashicorp` tooling.

PR #156, opened 2021-02-22
runs create-builder 5 times before failing

danielleadams:
AWS allows 1 unauthenticated pull per IP per second. This will hopefully throttle the pulls enough to have passing builds on the nightlys. There are 2 API endpoints that are being hit, BatchGetImage and GetDownloadUrlForLayer, that need to pass for a successful pull.


jabrown85 (approved):
> AWS allows 1 unauthenticated pull per IP per second

ಠ_ಠ

I very much disliked this, but I do like passing builds...so 🤷

jabrown85 (approved):

PR #157, opened 2021-02-25
Added builder descriptions

jabrown85:
This description is used by `pack` and other tooling. It is described in an upcoming RFC [here](https://github.com/buildpacks/spec/pull/193/files#diff-f42ff52873ec233ac5f2927cbd0d49bdb8f8c88e36198dadd42c73e2fef72f7dR77-R81).

### DIFF
--- a/builder-18.toml
+++ b/builder-18.toml
@@ -1,3 +1,5 @@
+description = "Base builder for Heroku-18 stack, based on ubuntu:18.04 base image"
+
 [stack]
 id = "heroku-18"
 build-image = "heroku/pack:18-build"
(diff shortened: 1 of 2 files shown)


hone (approved):

builder-18.toml
```diff
@@ -1,3 +1,5 @@
+description = "Base builder for Heroku-18 stack, based on ubuntu:18.04 base image"
```
joshwlewis, line 1: Should it say something about "Heroku supported buildpacks" or otherwise what buildpacks it contains?

joshwlewis (approved):

builder-18.toml
jabrown85: The buildpacks are listed in the metadata blob already. We could mention _something_, but it feels like a fine place to start.

PR #161, opened 2021-03-09
update heroku/nodejs versions

danielleadams:
Updates the Node.js buildpacks and removes the referenced buildpacks that are no longer needed in favor of the meta buildpack. Also upgrades the `procfile-cnb` versions for all the referenced buildpack groups in the builders.


jabrown85 (approved):
image

This looks great. The java buildpack is still pinned to `0.6.1`, but as long as there is no longer a conflict we are OK with that.

PR #162, opened 2021-03-10
Revert "Update the spring-boot builder and the makefile target"

jabrown85:
Reverts heroku/pack-images#144

`master` branch has not succeeded since. The error we are seeing is:

```
ERROR: failed to write image to the following tags: [heroku/buildpacks:20: image load 'heroku/buildpacks:20'. first error: embedded daemon response: invalid diffID for layer 27: expected "sha256:57a49fb084ace8d18b8b39aa70204f1e5f3b1ec979549e0805e63f11dd101e68", got "sha256:a02d3fde2aeb174a2aa54081c66d4ca7c61859ee120516d71817089c8236fa2e"]
```

### DIFF
--- a/spring-boot-builder-18.toml
+++ b/spring-boot-builder-18.toml
@@ -8,7 +8,7 @@ version = "0.10.1"
 
 [[buildpacks]]
   id = "heroku/jvm"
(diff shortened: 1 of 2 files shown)


jabrown85:
This did not work.

PR #164, opened 2021-03-31
Push produced images to another registry in addition to dockerhub.

jabrown85:
Push produced images to another registry in addition to dockerhub.

Blockers:
- [x] Make sure team is happy with `heroku/cnb/heroku:{stack}` pattern
- [x] Create registry repo
- [x] Setup CircleCI secrets
- [x] Ensure CircleCI does not allow secrets in forks

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -149,17 +149,36 @@ jobs:
       image_tag_alias:
         description: "Name of dockerhub image alias to publish"
         type: string
+        default: ""
+      image_tag_alias2:
+        description: "Name of dockerhub image alias to publish"
+        type: string
+        default: ""
     docker:
       - image: circleci/golang:1.15
     steps:
       - setup_remote_docker
       - run: docker login -u $DOCKER_HUB_USER -p $DOCKER_HUB_PASS
+      - run:
+          name: Log into additional registry
+          command: |
+            REGISTRY_TOKEN=$(curl -f -X POST $SERVICE_TOKEN_ENDPOINT -d "{\"username\":\"$SERVICE_TOKEN_USER_NAME\", \"password\":\"$SERVICE_TOKEN_PASSWORD\"}" -s --retry 3 | jq -r ".raw_id_token")
+            echo "$REGISTRY_TOKEN" | docker login $REGISTRY_HOST -u "$REGISTRY_USER" --password-stdin
       - attach_workspace:
           at: /tmp/workspace
       - run: docker load < /tmp/workspace/<< parameters.image_file >>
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -316,7 +335,7 @@ workflows:
           name: publish-18-build-stack
           image_file: pack-18-build.tar
           image_tag: heroku/pack:18-build
-          image_tag_alias: heroku/pack:18-build
+          image_tag_alias: $REGISTRY_HOST/s/heroku/cnb/heroku:18-build
```
jsncmgs1, line 338: What's the `/s/` in these aliases? stack?

jsncmgs1 (approved):

.circleci/config.yml
joshwlewis: Since we all have`/u/`, for our own accounts I'm thinking it's short for user? Maybe `/s/` is service? Totally guessing 🤷🏻

.circleci/config.yml
```diff
@@ -333,7 +352,7 @@ workflows:
           name: publish-18-run-stack
           image_file: pack-18-run.tar
           image_tag: heroku/pack:18
-          image_tag_alias: heroku/pack:18
+          image_tag_alias: $REGISTRY_HOST/s/heroku/cnb/heroku:18
```
joshwlewis, line 355: Should this be `cnb/pack:18`? This isn't exactly `heroku:18`; it's got a few layers on top...

joshwlewis: Or maybe call it `heroku/cnb/heroku-run:18` so that it's disambiguated from `heroku/heroku:18` and hints at it's purpose in CNB land?

jabrown85: yeah, I always disliked `pack:18` naming. My thought was that since this is under the `cnb` umbrella repository we could ditch the `pack` in favor of `heroku`. We could eventually have `cnb/bionic` and `cnb/functions:20` and whatnot.

jabrown85: I'm sort of partial to only having `-build` in the tag. Since, ultimately, customer workloads run on the `cnb/heroku:20` image. It's the most "public" version and I like not having the prefix. I could see us wanting to reference that image name if we allow custom images that derive from that image for rebase-ability or something. But I don't feel too strongly, we can always move things around.

hone: I'm 👍 on the new image changes:
* `heroku/cnb/heroku-18:build` - Build
* `heroku/cnb/heroku-18:run` - Run
* `heroku/cnb/heroku-18:builder` - CNB Builder

hone (approved):

PR #167, opened 2021-04-23
heroku/nodejs-function has moved

dzuelke:
https://github.com/heroku/buildpacks-nodejs/pull/42


dzuelke:
(mot published yet, so won't pass)

Malax:
Can you use hashes like we do with the others? We actually had an issue before when using tags.

edmorley:
@Malax I think that should be fixed by buildpacks/pack/pull/1111

PR #192, opened 2021-08-10
Remove the evergreen/fn meta-buildpack

edmorley:
Since:
* Nothing currently uses it.
* It's another layer of indirection that makes understanding the builder harder, as concepts are split across files.
* The Evergreen name is an obsolete codename.

If in the future we find it useful to have a way to group together functions buildpacks, we can always add another (more-suitably) named meta-buildpack back. However if doing that, we should likely also do the same for non-functions too, and fully document it.

Fixes GUS-W-9732440.


edmorley:
After this lands I'll also update the docs here:
https://github.com/heroku/languages-team/blob/main/languages/cnb/deploy.md

PR #214, opened 2021-10-28
Spike on automating pack-image file updates

schneems:
Spike on automating pack-image file updates
The problem I'm addressing is somewhat described in https://salesforce.quip.com/J3ZpAihiOAaZ. Currently, deploying to `pack-images` is manual. Developers must deploy the changes and then manually update the builder files.

As seen in https://github.com/heroku/pack-images/pull/212, the dev must remember to edit two files and two separate locations (with the same info).

This PR spikes out automating updating the files, which could be extended out to automate the entire deploy via a GitHub action eventually.

## Difficult to get right

To prove my point that it's difficult and error-prone to update these files manually when I ran the script, it showed that:

- heroku/procfile buildpack is wildly out of date
- heroku/scala is out of date
- heroku/nodejs is out of date

(See changes in builder-18.toml, and builder-20.toml files)

## Automation

This change works by reading in the existing files, querying the "registry" (just a GitHub repo) https://github.com/buildpacks/registry-index, and then updating the URI and version numbers on disk to match the last non-yanked registry entry for that buildpack.

## Use instructions

```
$ bundle exec rake fetch:latest
```

## Concerns

- Machine versus human toml

The `TomlRB` gem currently does not "pretty" print or format the toml output in any way. That means our indentation and logical newlines will be different between humans and machines updating this file.

While it's jarring for now, if mostly machines update this file, humans will eventually not care about the formatting.

If it's a huge concern, we could upstream a "pretty print" mode to the toml-rb gem, but that may be a significant effort.

- All latest all the time

This code assumes we want the latest versions always. Since the script still needs to be manually executed and the developer still needs to commit and make a PR. The PR must be approved...there's still an opportunity to manually point at older versions, but it won't be the default.

- Doesn't handle shimmed buildpacks

Currently, the PHP, Python, and GO buildpacks use a shim instead of the registry. I have to skip them manually.

## Definition of Done

For a definition of done, I need feedback on the above concerns.

- Add unit tests
- Port logic out of the Rake task and towards testable classes
- Update the README and docs to reflect the new workflow
- Handle shimmed buildpacks
- Take this as an opportunity to talk about how we would get this running on GitHub-actions along with signed commits

GUS-W-10100382

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,0 +1,5 @@
+source "https://rubygems.org"
+
+gem "toml-rb"
+gem "excon"
+gem "rake"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,0 +1,19 @@
+GEM
+  remote: https://rubygems.org/
+  specs:
+    citrus (3.0.2)
+    excon (0.87.0)
+    rake (13.0.6)
+    toml-rb (2.0.1)
+      citrus (~> 3.0, > 3.0)
+
+PLATFORMS
+  x86_64-darwin-20
+
+DEPENDENCIES
+  excon
+  rake
+  toml-rb
+
+BUNDLED WITH
+   2.2.27
--- a/Rakefile
+++ b/Rakefile
@@ -1,0 +1,133 @@
+require "pathname"
+require "toml-rb"
+require 'json'
+require 'excon'
+
+def root_dir
+  Pathname(__dir__)
+end
+
+class RegistryEntry
+  attr_reader :ns, :name, :version, :yanked, :addr
+
+  def initialize(ns: , name: , version: , yanked: , addr:)
+    @ns = ns
+    @name = name
+    @version = version
+    @yanked = yanked
+    @addr = addr
+  end
+
+  def uri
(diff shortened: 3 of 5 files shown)


edmorley (changes requested):
Thank you for looking into this, and sorry for the delayed review!

I absolutely agree we should make the process of rolling out buildpack updates easier/more automated, however I'm not sure this is the approach we should pick. 

With a pull based approach like this, if a buildpack is intentionally not releasing updates yet (eg blocked on something or had rolled back), then any time this script is run it will pull in all updates.

Instead I think we should use a push based approach using GitHub actions event dispatch. That way when a buildpack release occurs, that repo pings this one, triggering a GitHub action that updates just that buildpack's metadata here.

I also don't think we should add any more Ruby scripts to repositories that don't use Ruby, given our transition to Rust. It would be ideal to consolidate around as few languages as possible (eg Rust, bash scripts).

edmorley:
Closing due to inactivity, conflicts, and since I think we'll want to pick a different approach here (more on this above).

PR #235, opened 2022-04-04
Add heroku:22 images and automation

joshwlewis:
This introduces  build, run, and builder images and automation based on the `heroku:22` stack image. This will make it possible to run CNB builds on the new `heroku-22` stack. 

`heroku/buildpacks:22` is in development and will start out very basic, with only buildpacks known to be working on `heroku-22` included. Other language buildpacks will be added as they gain support for `heroku-22`.

[W-10345122](https://gus.my.salesforce.com/one/one.app#/alohaRedirect/a07EE00000ZZE8oYAH)

### DIFF
--- a/builder-22.toml
+++ b/builder-22.toml
@@ -9,10 +9,6 @@ run-image = "heroku/pack:22"
 version = "0.13.5"
 
 [[buildpacks]]
-  id = "heroku/ruby"
-  uri = "docker://public.ecr.aws/heroku-buildpacks/heroku-ruby-buildpack@sha256:3eeb3773cdbd29d4fb0d578f7781fe8c525de73593480e2740b7143262e5bef5"
-
-[[buildpacks]]
   id = "heroku/procfile"
   uri = "docker://docker.io/heroku/procfile-cnb:1.0.1"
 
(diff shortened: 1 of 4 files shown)


builder-22.toml
```diff
@@ -0,0 +1,31 @@
...
+  uri = "docker://public.ecr.aws/heroku-buildpacks/heroku-nodejs-buildpack@sha256:c4de76730ea8d0c68ccf1c073734ae467890545eec5896fa5ae805e2883ff20d"
+
+[[buildpacks]]
+  id = "heroku/nodejs-function"
+  uri = "docker://public.ecr.aws/heroku-buildpacks/heroku-nodejs-function-buildpack@sha256:6989739efd5c866907bedf235e0e350bb53938d482377c1bf116f70752794f41"
+
+[[order]]
+  [[order.group]]
+    id = "heroku/nodejs-function"
+    version = "0.9.1"
```
edmorley, line 26: This buildpack version and the one below differs from that in the heroku-20 builder. That is quite possibly expected (since the most recent changes to the nodejs CNB were just to add Heroku-22 support, which the other builders don't need), but mentioning just in case it wasn't :-)

edmorley (approved):
Thank you for doing this!

I really like the new table in the readme. 

I wonder if we should also add a sentence (not necessarily in this PR, but I don't mind either way) saying that the CNBs other than Node/Java are a WIP and subject to breaking changes at any time? (eg for when the shimmed buildpacks are swapped out?)

builder-22.toml
joshwlewis: Yes, this was intentional. I wasn't sure how long this branch would hang around, so was avoiding updating `builder-18.toml` and `builder-20.toml` to prevent potential merge conflicts later.

PR #251, opened 2022-06-03
Introduce builder-22 and builder-classic-22

joshwlewis:
To help with builder discovery and understanding this PR:

- Adds `heroku/builder:22`: A heroku-22 builder that includes Heroku's first-party, native Cloud Native Buildpacks.
- Adds `heroku/builder-classic:22`: A heroku-22 builder that includes Heroku's first-party classic buildpacks, shimmed for compatibility with Cloud Native Buildpacks APIs.
- Drops `heroku/buildpacks:22`: Previous `heroku/buildpacks:*` were a mix of classic/shimmed and native buildpacks that won't be repeated for 22 and greater.

GUS-W-11227981.


joshwlewis:
> After this lands, what do you think about renaming `pack-images` to `builder-images` or similar?

Yes, I'd like to do that. `heroku/builder-images` makes sense. I've also considered `heroku/builder`; I might like it better.

edmorley:
Either work for me :-)

edmorley:
It looks like more output can be generated if `HEROKU_DEBUG_RAILS_RUNNER` is set:
https://github.com/heroku/heroku-buildpack-ruby/blob/fe4db44101a9482c09cadd6e5b5ebf112e278f79/lib/language_pack/helpers/rails_runner.rb#L74-L140

But another thing that is strange, is that if the command fails, there's supposed to at least be this output (but there is not)?

```
Detecting rails configuration failed
set HEROKU_DEBUG_RAILS_RUNNER=1 to debug
```

joshwlewis:
With `pack build rgs --builder heroku/builder-classic:22 --trust-builder --env "HEROKU_DEBUG_RAILS_RUNNER=1"` I see the following logs:

```
-----> Detecting rails configuration
       $ rails runner "begin; puts %Q{heroku.detecting.config.for.assets.compile=#{Rails.application.config.try(:assets).try(:compile)}}; rescue => e; puts e; puts e.backtrace; end; begin; puts %Q{heroku.detecting.config.for.action_dispatch.x_sendfile_header=#{Rails.application.config.try(:action_dispatch).try(:x_sendfile_header)}}; rescue => e; puts e; puts e.backtrace; end;"
       heroku.detecting.config.for.assets.compile=false
       heroku.detecting.config.for.action_dispatch.x_sendfile_header=
-----> Detecting rake tasks


exit status 127
ERROR: failed to build: exit status 3
ERROR: failed to build: executing lifecycle: failed with status code: 51
```

So, it looks like the failure might be coming from the rake task detection?

edmorley:
The failing step is now this one:
https://github.com/heroku/heroku-buildpack-ruby/blob/cdc7902735b512ade68561de752f31c611f39ba0/lib/language_pack/ruby.rb#L965-L968

So it's interesting that enabling debug logging meant it seemed the buildpack got further this time? (The `Detecting rake tasks` doesn't come from the step that was seeming to fail before.)

Perhaps the Ruby buildpack is buffering output more than it should, and the `Detecting rake tasks` output was previously caught in the buffer?

edmorley:
Summary of the issue encountered with Ruby:
1. Heroku-22 no longer ships with system Ruby (by design)
2. The Ruby buildpack's `bin/release` requires Ruby, but the Ruby installed for the customer doesn't end up on PATH (I haven't checked at this point if it's a limitation of `bin/release` or something the buildpack should be doing to get correct PATH there)
3. This failure is silently ignored on Heroku, since Codon ignores `bin/release` failures
4. It fails with shimmed CNBs, since cnb-shim doesn't ignore `bin/release` failures
5. There may also be a buffering/logging bug with the Ruby buildpack, since both on Heroku and also with the shimmed CNB, there is output missing in some cases
6. cnb-shim could probably also be doing more to debug (eg printing "bin/release exited with status X" etc)
7. For now the Ruby buildpack has switched to using bash for `bin/release`, in https://github.com/heroku/heroku-buildpack-ruby/pull/1308

.circleci/config.yml
```diff
@@ -57,9 +54,9 @@ jobs:
           at: /tmp/workspace
       - run: docker pull heroku/heroku:<< parameters.stack-tag>>-cnb
       - run: docker pull heroku/heroku:<< parameters.stack-tag>>-cnb-build
-      - run: docker load < /tmp/workspace/buildpacks-<< parameters.stack-tag >>.tar
-      - run: pack trust-builder heroku/buildpacks:<< parameters.stack-tag >>
-      - run: pack build pack-getting-started --path getting_started --builder heroku/buildpacks:<< parameters.stack-tag >> --pull-policy never
+      - run: docker load < /tmp/workspace/<< parameters.builder-dir >>.tar
+      - run: pack trust-builder << parameters.builder-dir >>
```
edmorley, line 58: (Optional) In the CI run I see a (pre-existing) deprecation warning:

> Warning: Command pack trust-builder has been deprecated, please use pack config trusted-builders add instead

builder-classic-22/builder.toml
```diff
@@ -0,0 +1,111 @@
...
+[stack]
+id = "heroku-22"
+build-image = "heroku/heroku:22-cnb-build"
+run-image = "heroku/heroku:22-cnb"
+
+[lifecycle]
+version = "0.14.0"
+
+[[buildpacks]]
+  id = "heroku/java"
```
edmorley, line 12: Should these shimmed CNBs have different IDs to their non-shimmed counterparts?

eg:
```toml
  id = "heroku/java-classic"
```

builder-classic-22/builder.toml
```diff
@@ -0,0 +1,111 @@
...
+  uri = "https://cnb-shim.herokuapp.com/v1/heroku/java?version=0.3.1&name=Java"
```
edmorley, line 13: I presume the `version` param here is mostly redundant, since it doesn't affect the returned buildpack other than what's put into `buildpack.toml` - since the latest stable release is always returned?

If so, should we just zero the version out or something, to indicate it's not actually real?

Longer term I guess we could try and make cnb-shim fetch the version from the original buildpack registry, for shims using the alias naming scheme, if we actually cared about the version?

builder-classic-22/builder.toml
```diff
@@ -0,0 +1,111 @@
...
+
+[[buildpacks]]
+  id = "heroku/ruby"
+  uri = "https://cnb-shim.herokuapp.com/v1/heroku/ruby?version=0.3.1&name=Ruby"
+
+[[buildpacks]]
+  id = "heroku/scala"
+  uri = "https://cnb-shim.herokuapp.com/v1/heroku/scala?version=0.3.1&name=Scala"
+
+[[order]]
```
edmorley, line 43: Should this order match the default Heroku detection order?

If so, that order is:
`ruby clojure python java gradle scala php go nodejs`

(from https://github.com/heroku/codon/blob/dd40cf73232a39e55279a9f03bdcab9c924ecc32/lib/codon/buildpacks.rb#L9)

README.md
```diff
@@ -1,27 +1,33 @@
...
-| [heroku/pack:22][pack-tags]             | [heroku/heroku:22][stack-images]       | CNB Run Image     | Deprecated (replaced by [heroku/heroku:22-cnb][heroku-tags])           |
-| [heroku/pack:22-build][pack-tags]       | [heroku/heroku:22-build][stack-images] | CNB Build Image   | Deprecated (replaced by [heroku/heroku:22-cnb-build][heroku-tags])           |
-| [heroku/buildpacks:22][buildpacks-tags] | [heroku/heroku:22-build][stack-images] | CNB Builder Image | In Development                                                         |
+This repository is responsible for building and publishing [Cloud Native Buildpacks](https://buildpacks.io)
+builders that enable Heroku-like builds with the [`pack`](https://github.com/buildpacks/pack) command.
+
+| Builder Image                                       | Base Image                        | Status         |
+|-----------------------------------------------------|-----------------------------------|----------------|
+| [`heroku/buildpacks:18`][buildpacks-tags]           | [`heroku/heroku:18`][heroku-tags] | Available      |
```
edmorley, line 10: Should these have the base images listed as the `-cnb-build` variants? (Though I guess technically they have two base images, the build and the runtime, but shrug :-) )

builder-classic-22/builder.toml
joshwlewis: According to https://github.com/heroku/pack-images/pull/44, these version numbers are/were meaningful for cache busting. I don't know which cache that this busts, but if that's true, we'd quickly increment past `0.0.0`.

The existing numbers were copied from `heroku/buildpacks:20`.

builder-classic-22/builder.toml
joshwlewis: This order was ported from `heroku/buildpacks:20`, but yeah, I think the platform order makes more sense.

README.md
joshwlewis: Yeah, makes sense. Done.

builder-classic-22/builder.toml
joshwlewis: I tried; it didn't work. The ids here need to match the id of the downloaded buildpack. If we changed the `id` here to `heroku/java-classic`, we'd need to change it in the shim to match. If this is something we want (which I'm not opposed to), I'd rather do it in a supplemental PR.

edmorley: > The ids here need to match the id of the downloaded buildpack.

Ah I didn't realise. In which case yeah agree we should defer (if we decide to even do it).

builder-classic-22/builder.toml
edmorley: @joshwlewis I can't see any caching in cnb-shim's web server:
https://github.com/heroku/cnb-shim/blob/main/server/main.go

...nor any mention of "cache" in the git history for cnb-shim.

Looking at the pack-images repo state at the time of #44, I see caching in the Circle CI config that may have been the issue here?
https://github.com/heroku/pack-images/blob/1a646e80102df84bd5440d11209cb0c49e9bead0/.circleci/config.yml#L19-L22

> The existing numbers were copied from heroku/buildpacks:20.

Yeah I was just thinking that since this is a brand new image with "parity with Heroku" in mind, we're kind starting with a clean slate :-)

I'm happy either way though.

edmorley:
@joshwlewis Do we want to delete the superseded `heroku/buildpacks:22` image tag now, or wait a bit? If people have started using it (even though it was missing all but Node CNBs and experimental), I'm not sure that waiting longer will necessary increase the chance they've migrated?

https://hub.docker.com/repository/docker/heroku/buildpacks/tags?page=1&ordering=last_updated&name=22

joshwlewis:
> Do we want to delete the superseded heroku/buildpacks:22 image tag now, or wait a bit?

It's done.

PR #252, opened 2022-06-04
Remove redundant `create-docker-config` CI step

edmorley:
Since it's no longer required for CI to pass (I couldn't find the exact reason why it was added, since it was undocumented, but I believe Pack used to not handle a missing config well).


joshwlewis (approved):
Thanks. I found myself wondering about this step too. I'm sure there was a reason once upon a time!

PR #255, opened 2022-06-08
Switch from Circle CI's remote Docker feature to Machine

edmorley:
Previously:
- There was inconsistency between jobs, with some using Circle CI's remote Docker feature, and others using `machine` (full VM).
- For the jobs using remote Docker, they were using an outdated and insecure version of Docker, since Circle CI defaults to an ancient Docker version, unless one is explicitly specified.

Rather than specifying a more up to date remote Docker version, I've switched all jobs to using `machine`, since it's faster for our use-case (particularly since we `docker {save,load}` which requires passing bits over the network to the under-powered remote Docker instance). In addition, there are often delays in the `"Setting up a remote Docker instance"` step, negating any time savings using remote Docker over a full `machine` VM.

As an added bonus, the Ubuntu 22.04 `machine` images ship with zstandard installed by default (unlike `cimg/base` or the Ubuntu 20.04 machine image), which allows us to compress the exported builder tarfile for faster workspace persistence.

The `resource_class` definitions have been removed, since testing with larger instance sizes has shown they provide negligible benefit in terms of build/test times, and in fact make end to end times worse due to having longer average `"preparing instance"` times. (I would presume this is due to Circle CI having larger pre-warmed VM pools for default sizes than custom sizes.)

These changes reduce the "build + test" end to end times for PRs **from ~6.5 minutes to ~3.5 minutes**.

See:
https://circleci.com/docs/2.0/building-docker-images/
https://circleci.com/docs/2.0/configuration-reference/#machine
https://circleci.com/developer/machine/image/ubuntu-2204
https://circleci.com/docs/2.0/configuration-reference/#resourceclass
https://github.com/facebook/zstd

GUS-W-11263169.


joshwlewis (approved):
Seems fast!

edmorley:
Auto-merge wasn't working, and after disabling, it shows:
> The base branch requires all commits to be signed.

edmorley:
(Presume it was GitHub's rebase button thing that doesn't sign)

joshwlewis:
> Presume it was GitHub's rebase button thing that doesn't sign

Yeah, it does that. Do we need this signed commit branch protection?

edmorley:
I'm fine with turning it off

PR #260, opened 2022-06-09
Update JVM buildpacks

Malax:
## `heroku/java` `0.6.0`

* Upgraded `heroku/maven` to `1.0.1`
* Upgraded `heroku/jvm` to `1.0.1`

### Breaking
* Remove Gradle support from this meta-buildpack. Gradle support was realized by using a shimmed version of the `heroku/gradle` Heroku buildpack. We decided to strictly separate shimmed buildpacks from proper CNBs. Gradle support will be re-added later, using a native CNB. ([#308](https://github.com/heroku/buildpacks-jvm/pull/308))

## `heroku/java-function` `0.3.31`

* Upgraded `heroku/jvm-function-invoker` to `0.6.2`
* Upgraded `heroku/maven` to `1.0.1`
* Upgraded `heroku/jvm` to `1.0.1`

## `heroku/jvm` `1.0.1`

* Add support for the `heroku-22` stack. ([#304](https://github.com/heroku/buildpacks-jvm/pull/304))
* [Azul Zulu Builds of OpenJDK](https://www.azul.com/downloads/?package=jdk#download-openjdk) is now the default OpenJDK distribution. This change does not affect the `heroku-18` and `heroku-20` stack. ([#304](https://github.com/heroku/buildpacks-jvm/pull/304))

## `heroku/maven` `1.0.1`

* Add support for the `heroku-22` stack. ([#304](https://github.com/heroku/buildpacks-jvm/pull/304))

## `heroku/jvm-function-invoker` `0.6.2`

* Upgrade `libcnb` to `0.6.0` and `libherokubuildpack` to `0.6.0`.
* Add support for the `heroku-22` stack. ([#304](https://github.com/heroku/buildpacks-jvm/pull/304))

### DIFF
--- a/builder-22/builder.toml
+++ b/builder-22/builder.toml
@@ -16,6 +16,14 @@ version = "0.14.0"
   id = "heroku/nodejs-function"
   uri = "docker://public.ecr.aws/heroku-buildpacks/heroku-nodejs-function-buildpack@sha256:7c7e4453ff62639bd2044001c0f504092c53ecff5b3a720adef7866611e5dea7"
 
+[[buildpacks]]
+  id = "heroku/java"
+  uri = "docker://public.ecr.aws/heroku-buildpacks/heroku-java-buildpack@sha256:e3031bc05816379243c5f732e4acb4345f7c968977b91bae0e40dd954894b87e"
+
+[[buildpacks]]
+  id = "heroku/java-function"
+  uri = "docker://public.ecr.aws/heroku-buildpacks/heroku-java-function-buildpack@sha256:bd64e489115813777d8723f04944a98fcf7993329d55db6c6c80a9b1c05599d0"
+
 [[order]]
   [[order.group]]
     id = "heroku/nodejs-function"
@@ -25,3 +33,13 @@ version = "0.14.0"
   [[order.group]]
     id = "heroku/nodejs"
     version = "0.5.4"
+
+[[order]]
+  [[order.group]]
+    id = "heroku/java-function"
+    version = "0.3.31"
+
+[[order]]
+  [[order.group]]
+    id = "heroku/java"
+    version = "0.6.0"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -170,7 +170,7 @@ workflows:
           matrix:
             alias: test-getting-started-guides-22
             parameters:
-              language: ["node-js", "typescript"]
+              language: ["java", "node-js", "typescript"]
           requires:
             - create-builder-22
       - test-example:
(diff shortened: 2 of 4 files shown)


joshwlewis (changes requested):
Would you mind adding integration tests in the circle config for heroku/builder:22? It's currently only testing Node.js, Typescript, Node.js functions, and Typescript functions.

Malax:
@joshwlewis Of course, let's do it! :)

builder-22/builder.toml
```diff
@@ -25,3 +33,13 @@ version = "0.14.0"
   [[order.group]]
     id = "heroku/nodejs"
     version = "0.5.4"
+
+[[order]]
+  [[order.group]]
+    id = "heroku/java-function"
```
edmorley, line 39: What detection order should we use for these?

With the PR as is, we'll have three different orders in use in our builders (even just considering the Java/Node.js buildpacks and ignoring the other buildpacks for a moment):

- `heroku/buildpacks:{18,20}`:
  -  ..., `heroku/nodejs-function`, `heroku/java-function`, `heroku/nodejs`, `heroku/java`
- `heroku/builder-classic:22`:
  -  ..., `heroku/java`, ..., `heroku/nodejs`
- `heroku/builder:22` (this PR):
  - `heroku/nodejs-function`, `heroku/nodejs`, `heroku/java-function`, `heroku/java`

Malax: Yeah, it's inconsistent for sure. For the changes in this PR, I went with "copy what's already there" and replicated the order of the Node.js stuff. But since there was only one language in the builder, it's hard to derive intent from that. :)

I think the function buildpacks always need to come first since they would build just fine with the regular language buildpacks too - just without the language runtimes.

That leaves us with making decision how we order the languages themselves. Unless there is a technical reason for a specific order, alphabetical seems easy enough.

Malax:
@joshwlewis Is 7e0dd2a what you had in mind?

joshwlewis (approved):

Malax:
@joshwlewis No Scala and Clojure CNBs yet. :(

builder-22/builder.toml
edmorley: @Malax The order can have a significant impact for projects that can match against multiple languages. The chance of this is higher for node.js where you'll get projects with both a front-end and backend component.

As to what order we should actually use, it's not clear :-)

Malax: @edmorley That sounds like this is not a blocker for this PR? Where should we move this if at all?

PR #261, opened 2022-06-10
Remove retrying of `pack builder create`

edmorley:
Since:
- The retry handling is broken, such that if the command fails after three retries, then the exit code is obscured and the CI run continues to the next step. This causes a confusing error message that's shown instead of the actual error.
- In the case of legitimate failures, the retries prolong the time until the CI result shows the failure.
- The builder creation step is pretty reliable now (I've inspected several dozen invocations and didn't see any failures).
- If we start seeing intermittency in the future, I would rather know about it so we can improve reliability for everyone, rather than cover it up via retries.

GUS-W-11263169.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,16 +16,7 @@ jobs:
           version: 0.26.0
       - attach_workspace:
           at: /tmp/workspace
-      - run:
-          name: Create builder with retries
-          command: |
-            n=1
-            until [ "$n" -ge 5 ]
-            do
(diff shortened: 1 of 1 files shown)


edmorley:
Ah true! :-) 

I'm presuming some other failure modes are errors or rate limits pulling from Docker Hub and/or ECR. However Circle CI now has exemption from Docker Hub rate limits even when not authed (I think there was a spell before that was implemented), and the ECR rate limit doesn't seem to be an issue in practice (whilst the limit is low, it's per Docker repo per IP, and unlike the buildpack cutlass tests, this CI doesn't do repeat pulls of the same Docker repo within the same job).

PR #262, opened 2022-06-13
Make getting-started + functions tests actually test the server

edmorley:
Previously the getting started guide smoke tests, as well as the example functions tests both only checked that the app built and the default process' container ran long enough to `docker exec` into.

This didn't provide coverage of things like whether the default process was actually a web server that correctly binds to the provided `PORT`, or if it successfully returned an HTTP 200.

Now:
- The tests check for an HTTP 200 from the specified `PORT`.
- More information is output in the case of failure.
- Retries are used instead of a fixed 5 second sleep when waiting for the container to start.

In addition, the example functions test directories and Circle CI job names have been renamed to make them more readable/identifiable.

GUS-W-11263169.


.circleci/config.yml
```diff
@@ -36,22 +36,31 @@ jobs:
...
+          command: |
+            if curl -sSf --retry 5 --retry-delay 1 --retry-all-errors --connect-timeout 3 http://localhost:8080 -o response.txt; then
+              echo "Successful response from server"
+            else
+              echo "Server did not respond successfully"
+              docker logs getting-started
+              [[ -f response.txt ]] && cat response.txt
+              exit 1
+            fi
+  test-example-function:
```
edmorley, line 60: I've renamed this job type from `test-example` to `test-example-function`, because the `x-health-check` handling in the `Test function response` step is functions specific.

If we ever have additional fixtures under `examples/` that are no functions, we can add a new job type at that point to test them separately.

Malax (approved):

PR #263, opened 2022-06-14
Update example functions to latest generated versions

edmorley:
The example functions used by tests in this repository are from the scaffolds generated by the SFDX CLI.

These examples have been regenerated by:
1. Installing the SFDX CLI:
  https://developer.salesforce.com/tools/sfdxcli
2. Creating a new Salesforce DX project:
  `sfdx force:project:create -n myproject`
3. Within that project, creating an example function for each language:
  ```
  sf generate function -n javafunctionscaffold -l java
  sf generate function -n javascriptfunctionscaffold -l javascript
  sf generate function -n typescriptfunctionscaffold -l typescript
  ```
4. Replacing the contents of `examples/*-function/` with the generated files.

GUS-W-11263169.


edmorley:
@joshwlewis As part of regenerating the latest scaffold functions for this PR, I noticed that the scaffolds are still using `salesforce-api-version` `53.0` even though the template was updated and `plugin-functions` bumped:
https://github.com/heroku/sf-functions-core/pull/46
https://github.com/salesforcecli/plugin-functions/pull/403

Investigating, it looks like those changes to `plugin-functions` have been released in `1.9.0` on 19th May:
https://github.com/salesforcecli/plugin-functions/tags

...but that it then needs a further manual bump in `cli` itself (😫):
https://github.com/salesforcecli/cli/blob/554823eab26c9d217c8143389f58a933dcf0eb3d/package.json#L102

This feels like something Dependabot could/should do? I'm not sure why it didn't open a PR:
https://github.com/salesforcecli/cli/blob/main/.github/dependabot.yml
(`monthly` is run on 1st of the month, so should already have run. And this wasn't a semver-major so shouldn't be covered by the exclude rule)

Would you mind bumping the dep in `cli` manually to release that change?

For now, I've manually bumped the templates in this PR to use `54.0`, to save having to update them again in a bit :-)

joshwlewis (approved):

PR #264, opened 2022-06-15
Reinstate Gradle support for `heroku/buildpacks:{18,20}`

Malax:
`heroku/java` previously supported Gradle by mixing in the shimmed `heroku/gradle` buildpack. When we decided to make a clean cut and not have shimmed buildpacks in the CNB repository, support for Gradle in `heroku/java` was dropped.

This is obviously a major breaking change and this PR reinstates Gradle support by mirroring the order definition that was previously in `heroku/java`. This is not a complete fix since users that explicitly used `heroku/java` (i.e. `--buildpack heroku/java` with `pack`) will still get detection failures for Gradle. However, using the builder without explicit buildpacks will work fine for Gradle projects.

Closes GUS-W-11298364

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -149,7 +149,7 @@ workflows:
           matrix:
             alias: test-getting-started-guides-18
             parameters:
-              language: ["go", "java", "node-js", "typescript", "php", "python", "ruby"]
+              language: ["go", "java", "node-js", "typescript", "php", "python", "ruby", "gradle"]
           requires:
             - create-buildpacks-18
       - test-getting-started-guide:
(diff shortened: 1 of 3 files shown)


edmorley (commented):
Could you add `gradle` to the getting-started-guide matrix here?
https://github.com/heroku/builder/blob/4710f39ecf2a7df98bff278fd9a3281a1b964579/.circleci/config.yml#L145-L164

.circleci/config.yml
```diff
@@ -149,7 +149,7 @@ workflows:
           matrix:
             alias: test-getting-started-guides-18
             parameters:
-              language: ["go", "java", "node-js", "typescript", "php", "python", "ruby"]
+              language: ["go", "java", "node-js", "typescript", "php", "python", "ruby", "gradle"]
```
edmorley, line 152: These were in alphabetical order, but fine with fixing that (and making the order consistent with that used for `heroku/builder-classic:22`'s test `matrix`) in the later PR to add `scala` to save another RTT

Malax: 👍🏻

PR #282, opened 2022-09-27
Use shimmed Ruby buildpack instead of the prototype hybrid CNB

schneems:
The previously deployed Ruby buildpack is based on the "hybrid" approach https://github.com/heroku/heroku-buildpack-ruby. Essentially Terence added support for CNB into the existing buildpack. This works, but it's not taking advantage of most of the features in CNB. It's really not too different than using the "shim" buildpack.

(The *shim* is a wrapper around any legacy/v2 buildpack that allows it to interface with CNB)

The idea is that we'll use the shim buildpack for now and switch over once a native CNB implementation in Rust is ready.

GUS-W-11461073

### DIFF
--- a/buildpacks-18/builder.toml
+++ b/buildpacks-18/builder.toml
@@ -26,7 +26,7 @@ version = "0.14.1"
 
 [[buildpacks]]
   id = "heroku/ruby"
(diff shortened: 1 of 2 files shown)


edmorley:
@schneems Looks like your Circle CI OAuth needs refreshing (sign in and out). I worked around for now by closing/reopening, so Circle used my token (as the actor of the reopening event).

edmorley (approved):

PR #286, opened 2022-09-29
Convert to GitHub Actions

joshwlewis:
This converts the build, test, and publish CI workflows from CircleCI to GitHub Actions.

Notable differences from the CircleCI implementation:

- We can now trigger this manually via workflow_dispatch if needed
- It no longer runs on weekends to prevent surprises when nobody is around
- The sequencing is a little different - in Circle, the example tests for heroku-22 might run before the build of heroku-20 completes. In GitHub Actions, we're not starting testing until all the builders are built. This simplified the workflow dependencies a bit.

[W-11830543](https://gus.my.salesforce.com/one/one.app#/alohaRedirect/a07EE000017WyBkYAK)


.github/workflows/build-test-publish.yml
```diff
@@ -0,0 +1,161 @@
...
+
+  create:
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        builder: ["buildpacks-18", "buildpacks-20", "builder-classic-22", "builder-22"]
+    steps:
+      - uses: actions/checkout@v3
+      - uses: buildpacks/github-actions/setup-pack@v4.8.1
+      - name: Setup builder cache
```
edmorley, line 24: I presume the use of the cache rather than the artifacts feature is due to the slowdowns mentioned on Slack? Would it be worth adding a comment explaining why we're not using what GitHub suggests for inter job persistence?

Also, one concern I had was about the potential for cache poisoning (eg from a PR), however it seems the jobs from other branches cannot write to the `main` branch, if I'm understanding this correctly (so we should be fine):
https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache

.github/workflows/build-test-publish.yml
```diff
@@ -0,0 +1,161 @@
+name: Build, Test, Publish
+
+on:
+  pull_request:
+  push:
+    branches: ["main"]
+  schedule:
+    - cron: '0 0 * * *'
```
edmorley, line 8: I wonder if longer term we should consider making the cron less frequent at some point, particularly now there is the ability to trigger a new workflow via the GitHub Actions interface?

Changes like CNB version bumps, lifecycle version increases will trigger a build as part of the PR merging.

The only things that will actually get updated on cron jobs are components that are unpinned, so:
- base image updates (eg after the stack image release, once every two weeks)
- classic buildpack changes (which will get pulled in automatically by cnb-shim from the stable release on buildpack registry)

Perhaps a first step might be to avoid the cron firing over the weekend (or even a Friday)?

Though I suppose this would also delay things like any buildpack registry rollbacks propagating into the builder too.

Anyway, I guess something for another PR :-)

edmorley (approved):
Thank you for working on this! This was one of our most complex Circle CI workflows, so I can imagine it was a pain to port!

Some additional steps I had to perform around the time of merging for other repos (in case it's of use):
- Pressing "stop building" on the Circle CI project settings, so as to stop the failing "missing config" job making it look like `main` was failing CI
- Adjusting the GitHub required CI checks setting (for this repo I guess it's not practical to add every permutation of job - perhaps just the changeling + Actions `publish` jobs can be ticked?)
- Updating OrgH to switch the CI of record from Circle CI to GitHub Actions (and waiting 90+ mins for it to propagate, so Changeling stops being mad, though it still didn't work entirely for Python so I had to use the `?show_form=1` trick to override changeling for the PR, sigh)
- Removing secrets from the Circle CI project

.github/workflows/build-test-publish.yml
joshwlewis: Yes, this turned out to be 4x faster. Cache persistence and restoration was around 30s, where the upload-artifact was over 2 minutes. Also artifacts are available after the job completes, which is not something we care about here. 

My impression was that artifacts were better suited for build reports or production binaries, and caches are better suited for intra-workflow persistence.

I'll add a comment.

.github/workflows/build-test-publish.yml
joshwlewis: > Perhaps a first step might be to avoid the cron firing over the weekend (or even a Friday)?

Seems fine to me. I'll add it.

> Though I suppose this would also delay things like any buildpack registry rollbacks propagating into the builder too.

I think in this scenario, I'd prefer to trigger the action manually rather than wait for the scheduled run anyway, regardless of the frequency.

edmorley:
@joshwlewis It looks like the `publish` job is being skipped on `main` too? Not sure why at first glance.

eg https://github.com/heroku/builder/actions/runs/3176302715/jobs/5175517309

edmorley:
Ah, the `github` context doesn't have a `branch` attribute, but it does have a `ref` attribute:
https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

So something like this should work:

```
if: github.ref == 'refs/heads/main'
```

xref:
https://github.com/orgs/community/discussions/26349
https://stackoverflow.com/q/58139406

However it also seems that the default `if` for a job with a `needs` is `success()` and so by defining our own, it might be overriding the behaviour that skips the job if the previous jobs failed. See:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-not-requiring-successful-dependent-jobs
https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions

That is, I wonder if we really need to use this:

```
if: success() && github.ref == 'refs/heads/main'
```

PR #290, opened 2022-10-26
Update node function buildpacks

colincasey:
Update the Node Function buildpacks:
- https://github.com/heroku/buildpacks-nodejs/pull/373

[W-11854639](https://gus.my.salesforce.com/a07EE000017vsgyYAA)

### DIFF
--- a/.github/workflows/build-test-publish.yml
+++ b/.github/workflows/build-test-publish.yml
@@ -108,7 +108,7 @@ jobs:
       - run: docker run --name example-function --detach -p 8080:8080 --env PORT=8080 example-function
       - name: Test example function web server response
         run: |
-          if curl -sSf --retry 10 --retry-delay 2 --retry-all-errors --connect-timeout 3 -X POST -H 'x-health-check: true' http://localhost:8080 -o response.txt; then
+          if curl -sSf --retry 10 --retry-delay 3 --retry-all-errors --connect-timeout 3 -X POST -H 'x-health-check: true' http://localhost:8080 -o response.txt; then
             echo "Successful response from function"
           else
             echo "Function did not respond successfully"


colincasey:
@edmorley was discussing the build failures with @joshwlewis earlier.  i had to bump the curl timeout on checking the example functions to get the validations to pass.  any concerns about the changes to the workflow here from either of you?

edmorley:
@colincasey From the log output of the failing runs (where the container logs are dumped before the test run exits), it seemed that the runtime package was being downloaded on the fly at container boot? 

Example:

```
curl: (52) Empty reply from server
curl: (52) Empty reply from server
Function did not respond successfully
npm WARN exec The following package was not found and will be installed: @heroku/sf-fx-runtime-nodejs@0.12.0
```
(from https://github.com/heroku/builder/actions/runs/3332522944/jobs/5513599196#step:10:26)

If so:
1. That would explain why the timeouts need bumping
2. Means there is a bug in the buildpack change, since I presume this wasn't intended? 

I don't think this should be deployed as is, since installing packages at runtime is (a) non-deterministic, (b) slows down dyno boot, (c) causes a boot time dependency on npmjs.org being up etc.

edmorley:
@colincasey I realise what changes are being deployed, and I agree keeping the test fixtures in this repo the same for now (to confirm things still works with older functions codebases) makes sense. However, there appears to be an actual bug here. Before, even for the "implicit" runtime dep, the package was still installed during the *build*. Now, it's being installed at *runtime*.

edmorley:
I presume the issue is that at runtime, the invoker is now being launched via `npx` (when it was not before). And npx tries to be helpful and download a package if it cannot find it. 

Looking at the *build* log, the package is being installed:

```
Checking for function file
Installing Node.js Function Invoker Runtime @heroku/sf-fx-runtime-nodejs@0.12.0
```

So this presumably means npx isn't finding that package at runtime somehow, and therefore is going off and downloading it again, hence this at *runtime*:

```
npm WARN exec The following package was not found and will be installed: @heroku/sf-fx-runtime-nodejs@0.12.0
```

edmorley:
> npx isn't finding that package at runtime somehow

Ah I believe this is the cause...

In the Functions CNB, for the "project doesn't have the invoker in its package.json" case, the invoker is installed **globally** into its own layer (which makes sense), here:
https://github.com/heroku/buildpacks-nodejs/blob/c904daefbbf5da5db99733efaae5f957691cb647/buildpacks/nodejs-function-invoker/src/layers/runtime.rs#L60-L61

However [npx's docs](https://docs.npmjs.com/cli/v9/commands/npx) say:

> If any requested packages are not present in the local project dependencies, then they are installed to a folder in the npm cache, which is added to the PATH environment variable in the executed process.

To me that reads as "even if the package is installed globally/elsewhere, npx will re-download it, if it's not in the local node_modules".

I'm presuming adding `--no` (which means "don't install") to the `npx` call might work, so long as npx is willing to not install but still use whatever's on PATH.

That said, might it just be easier to run the `sf-fx-runtime-nodejs` entrypoint script itself, and not use npx in the buildpack at all?

joshwlewis:
I tend to agree that this isn't going to work as-is. The functions runtime only gives a function around 15s to boot up. It looks like this timeout is something more like 30s. We shouldn't need to install the package at boot anyway, since we already installed it during the build. Seems like there is a bug where `npx` isn't finding the previously installed `@heroku/sf-fx-runtime-nodejs`. Let's address that in the `heroku/nodejs-function-invoker` buildpack.

PR #302, opened 2022-12-05
Add heroku/go buildpack to builder:22

joshwlewis:


builder-22/builder.toml
```diff
@@ -24,6 +24,10 @@ version = "0.14.1"
   id = "heroku/java-function"
   uri = "docker://public.ecr.aws/heroku-buildpacks/heroku-java-function-buildpack@sha256:46e353cadcd941baeb0f566d7ab7b1f13c18e96223c1143dc51b258638ab9d47"
 
+[[buildpacks]]
+  id = "heroku/go"
+  uri = "docker.io/heroku/buildpack-go@sha256:fdc270c4414dc636daf29e269a67f28bb1a8c4aee89974d80b90edbdc10ae8a2"
```
edmorley, line 29: (for parity with the other entries, and also since I think otherwise the Pack CLI uses fuzzy heuristics to work out what the URL refers to)

```suggestion
  uri = "docker://docker.io/heroku/buildpack-go@sha256:fdc270c4414dc636daf29e269a67f28bb1a8c4aee89974d80b90edbdc10ae8a2"
```

builder-22/builder.toml
```diff
@@ -43,3 +47,8 @@ version = "0.14.1"
   [[order.group]]
     id = "heroku/java"
     version = "0.6.5"
+
+[[order]]
```
edmorley, line 51: Should the procfile buildpack be added to the Go group? (Similar to what I'll be doing for the Python buildpack)

For the other languages, they include the procfile buildpack in their meta-buildpack. However since the Go buildpack is not a meta-buildpack, the procfile buildpack isn't included. 

Whilst the Go buildpack does set default process types (as can be seen here: as can be seen here: https://github.com/heroku/builder/actions/runs/3623151336/jobs/6108735078#step:8:29) - if end users have a `Procfile` (for example to add additional commands used by their app, to customise args for the web process etc), then it will silently not do anything.

For example, the Go getting started guide actually has a `Procfile` at the moment (although it's very similar to the auto-detected processes, with the exception that without the Procfile, the Go builldpack actually configures two process, a `web` and a redundant `go-getting-started` - whereas the Procfile would have just configured `web`):
https://github.com/heroku/go-getting-started/blob/main/Procfile

README.md
```diff
@@ -12,11 +12,11 @@ builders that enable Heroku-like builds with the [`pack`](https://github.com/bui
 | [`heroku/builder:22`][builder-tags]                 | [`heroku/heroku:22-cnb-build`][heroku-tags] | Suggested  |
 | [`heroku/builder-classic:22`][builder-classic-tags] | [`heroku/heroku:22-cnb-build`][heroku-tags] | Available  |
 
-[`heroku/builder`][builder-tags] builder images feature Heroku's native Cloud Native Buildpacks. These buildpacks are optimized and make use of many CNB features.
+[`heroku/builder`][builder-tags] builder images feature Heroku's native Cloud Native Buildpacks. These buildpacks are optimized and make use of many CNB features. These builder images support Go, Java, Node.js, and Typescript codebases.
```
edmorley, line 15: Thank you for updating these descriptions!

edmorley:
Post merge, the `Build, Test, Publish / test-guides (builder-22, go)` CI job will also need adding to the `main` branch required status checks list.

builder-22/builder.toml
joshwlewis: Yeah, I considered adding the procfile buildpack to this group, but wanted feedback from y'all. I was surprised we haven't already added an optional procfile buildpack for Node.js and Java already, so it felt odd to add it for only Go. Honestly, I'm in favor, and was considering a separate PR to add Procfile to all the groups in this builder.

edmorley:
Ci is failing due to:
```
bash: line 1: bin/go-getting-started: No such file or directory
```

I presume this is because the CNB (rightly) no longer puts the built assets into `<app_dir>/bin/`?

Looking at the classic buildpack, I see it adds the `bin/` dir to `PATH`:
https://github.com/heroku/heroku-buildpack-go/blob/31b359312899520b97d4194b1f97c7da87b674ee/bin/compile#L685

As such, presumably the best fix here would be to modify the getting started guide `Procfile` to use `web: getting-started-guide` (ie without the `bin/` prefix), which should work with both the classic buildpack and CNB?

It does mean that one of the things we'll need to tell user when it comes to build unification, is the need for them to remove any `bin/` prefixes too, but that seems fine.

joshwlewis:
Setting back to draft due to CI failures.

joshwlewis:
Fix to the getting started guide is here: https://github.com/heroku/go-getting-started/pull/37. Yes, it seems safe to drop the `bin/` prefix from the Procfile. It works for both the classic and cloud native buildpack that way.

PR #312, opened 2023-01-31
Disable matrix `fail-fast` on GitHub Actions

edmorley:
Since otherwise any other in-progress jobs get cancelled, making it hard to see whether failures in one job only affect one permutation of the matrix, or several.

See:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

### DIFF
--- a/.github/workflows/build-test-publish.yml
+++ b/.github/workflows/build-test-publish.yml
@@ -16,6 +16,7 @@ jobs:
   create:
     runs-on: ubuntu-22.04
     strategy:
+      fail-fast: false
       matrix:
         builder: ["buildpacks-18", "buildpacks-20", "builder-classic-22", "builder-22"]
     steps:
@@ -35,6 +36,7 @@ jobs:
     runs-on: ubuntu-22.04
     needs: create
     strategy:
+      fail-fast: false
       matrix:
         builder: ["buildpacks-18", "buildpacks-20", "builder-classic-22"]
         language: ["go", "gradle", "java", "node-js", "php", "python", "ruby", "scala", "typescript"]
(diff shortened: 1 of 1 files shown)


joshwlewis:
I'm kind of meh on this. The failing job was listed on the top of the list in the GitHub UI earlier today. It seems a bit wasteful to run 6+ other jobs - especially when we are seeing some rate limit issues against ECR, which this change may compound. I don't have a strong opinion, though.

edmorley:
I think the fail-fast behaviour exacerbates the rate limit issue though - for Richard's PR along I've had to retrigger 3-4 times due to a JVM rate limit issue, just to be able to see the unrelated Ruby buildpack build logs.

Also, by the time jobs are cancelled, quite often one has paid most of the rate limit cost (since the download is up front) - all it's helping is saving 30s of test run time later on, which is dwarfed by repeating the setup tasks.

Even with fail-fast disabled, there is still some cancelling behaviour left - I believe the examples/guides tests won't start running unless all of the `create` jobs have succeeded - so if eg a typo is made in `buildpack.toml` then we won't waste time/rate limit on all the other jobs.

(Longer term I'd also like us to move Node.js/Java to Docker Hub to avoid the ECR public repo rate limit issues too)

edmorley:
Example:
https://github.com/heroku/builder/actions/runs/4058601388

(see the "latest run" dropdown to see all of the previous cancelled runs)

Malax:
I think this is fine. I agree with Josh that it might be wasteful - but a) failing CI is rare-ish since we usually just update some URLs making the cost argument less strong IMO. And if failing fast causes two engineers to waste cycles when something is borked - that almost certainly more expensive.

Malax (approved):

PR #322, opened 2023-03-07
Downgrade node guide test to v16 for heroku-18 stack

colincasey:
The [node-js-getting started guide](https://github.com/heroku/node-js-getting-started) should be compatible with both Node 16.x and 18.x but the Heroku-18 stack only supports Node 16.x.  This change adds an conditional step to the test workflow to downgrade the engine specified in `package.json` from `18.x || 16.x` to just `16.x`.

See:
https://github.com/heroku/node-js-getting-started/pull/294#discussion_r1127817002

GUS-W-12652981.

### DIFF
--- a/.github/workflows/build-test-publish.yml
+++ b/.github/workflows/build-test-publish.yml
@@ -55,6 +55,9 @@ jobs:
         with:
           ref: main
           repository: heroku/${{ matrix.language }}-getting-started.git
+      - name: Modify package.json (fix for Heroku-18 stack)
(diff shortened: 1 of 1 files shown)


edmorley (approved):
Thank you!

joshwlewis (approved):
Should solve the problem, though I look forward to removing this after the heroku-18 EOL!

edmorley:
Only ~7 weeks until Heroku-18 EOL 🎉

PR #335, opened 2023-05-03
Pin the buildpack versions in `heroku/buildpacks:18`

edmorley:
Since the `heroku/buildpacks:18` CNB builder image is based upon the Heroku-18 stack, which [has reached end-of-life](https://devcenter.heroku.com/changelog-items/2583), and as such support will soon be removed from the stack from our various buildpacks.

We will soon be stopping the publishing of `heroku/buildpacks:18` image updates entirely, however, we expect a few more upstream security updates to Ubuntu 18.04 (until 31st May, Canonical's revised EOL date) which we wish to pick up in the meantime.

By pinning the buildpack versions of our shimmed buildpacks, it unblocks removing Heroku-18 support from them in the meantime.

To pin the version, we rely on the fact that the buildpack registry stores versioned archives of each buildpack release, with a `vNNN` suffix.

However, we also have to:
- Pass an explicit `id` now, to prevent IDs like `heroku/python-v232`
- Pass an explicit `stacks` list due to: heroku/cnb-shim#72

The latest version numbers of each shimmed buildpack were generated using:

```
$ for lang in scala gradle ruby python php go; do echo "${lang}-v$(h buildpacks:versions heroku/${lang} | head -n3 | tail -n1 | cut -d ' ' -f 2)"; done
scala-v96
gradle-v39
ruby-v271
python-v232
php-v230
go-v173
```

GUS-W-13143718.

### DIFF
--- a/buildpacks-18/builder.toml
+++ b/buildpacks-18/builder.toml
@@ -14,7 +14,7 @@ version = "0.16.1"
 
 [[buildpacks]]
   id = "heroku/scala"
-  uri = "https://cnb-shim.herokuapp.com/v1/heroku/scala?version=0.0.0&name=Scala"
+  uri = "https://cnb-shim.herokuapp.com/v1/heroku/scala-v96?id=heroku/scala&name=Scala&version=0.0.0&stacks=heroku-18"
 
 [[buildpacks]]
   id = "heroku/java-function"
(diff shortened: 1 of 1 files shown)


edmorley:
CI is failing with:

```
ERROR: failed to add buildpacks to builder: invalid buildpack: buildpack from URI 'https://cnb-shim.herokuapp.com/v1/heroku/scala-v96?version=0.0.0&name=Scala' has ID 'heroku/scala-v96' which does not match ID 'heroku/scala' from builder config
```

This is because `cnb-shim` infers the buildpack ID from the registry URL, and it's (understandably, given we're trying to trick the URLs) treating the `-v96` in the URL as part of the buildpack name and not the version.

To resolve the error we would have to either:
1. Switch the IDs that we call the buildpacks - which is potentially a breaking change (given people may be using eg `--buildpack builder:heroku/scala` or whatever that syntax is) plus kinda meh for UX in build logs.
2. Or, update `cnb-shim` so that it allows overriding the ID via a URL parameter (the same way it allows overriding other attributes), or else so that it knows to strip off any `-vNNN` suffix.

Option (2) seems marginally better, however, I'm not sure it's worth the effort given that we could also just stop publishing new updates to `heroku/buildpacks:18`, given that when builds occur a fresh run image will be pulled in. (So any updates to the run image itself will still be picked up, even if we're not updating the builder image.)

edmorley:
I've decided to go with option (2) after all:
https://github.com/heroku/cnb-shim/pull/71

Malax (approved):

PR #344, opened 2023-05-16
Stop publishing the `heroku/buildpacks:latest` tag

edmorley:
Since:
- It's not documented, or used by anything on our side.
- The `heroku/buildpacks` namespace is deprecated, so calling anything to do with it "latest" seems like it could confuse users.
- We don't publish `latest` for the new `heroku/builder` namespace either (which is probably the best decision for now, given that having such a tag means people could receive breaking changes with no notice).

I will delete the `latest` tag from Docker Hub once this is merged.

GUS-W-13213079.

### DIFF
--- a/.github/workflows/build-test-publish.yml
+++ b/.github/workflows/build-test-publish.yml
@@ -142,7 +142,6 @@ jobs:
         include:
           - builder: buildpacks-20
             tag_public: heroku/buildpacks:20
-            tag_alias: heroku/buildpacks:latest
             tag_private: heroku-20:builder
           - builder: builder-classic-22
             tag_public: heroku/builder-classic:22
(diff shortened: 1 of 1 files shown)


edmorley:
@joshwlewis Yes, it would break anyone doing that - that's the point. The `latest` isn't mandatory for repos - and we don't have one for `heroku/builder` (so `--builder heroku/builder` doesn't work either). However, we have never documented to use `heroku/buildpacks` without specifying a tag, so in practice this would only break users who have not been following the recommended builder tag. 

Also, even if we decided to add the `latest` tag to `heroku/builder` (something that we would need to discuss separately, but I think may not be a good idea), as mentioned in the PR description, I still think we shouldn't have the `latest` tag for `heroku/buildpacks` since it's a deprecated namespace, and `latest` isn't really "latest".

joshwlewis (approved):

edmorley:
I opened a PR against Dokku a few weeks ago, to stop it using the `latest` tag, and that change was released in Dokku 0.30.5 on 2023-05-26:
https://github.com/dokku/dokku/releases/tag/v0.30.5

Dokku users who happen to be using the (experimental) Dokku CNB build backend (as opposed to some of the other build backends) can either upgrade to Dokku 0.30.5+, or else specify the `heroku/buildpacks:20` builder (or one of the newer builders, see: https://github.com/heroku/builder#heroku-builder-images) explicitly using one of the approaches documented here:
https://dokku.com/docs/deployment/builders/cloud-native-buildpacks/#customizing-the-buildpack-stack-builder

As such, I've just now removed the `latest` tag from the `heroku/buildpacks` Docker Hub repo:
https://hub.docker.com/r/heroku/buildpacks/tags

PR #358, opened 2023-06-30
add PHP to builder-22

dzuelke:


edmorley:
The language list in the readme can now be updated to include PHP :-)
https://github.com/heroku/builder/blob/8589058356b7614eb3e8882f60be9fe7418d41c0/README.md?plain=1#L15

PR #380, opened 2023-09-14
Improve the naming of the functions related CI jobs and test fixtures

edmorley:
* The CI job "test-examples" was already functions-specific, but is now renamed to "test-functions" to make that clearer.
* The functions test fixtures have been grouped together in a functions directory.

This change should make it slightly easier to understand the CI job matrix from the GitHub Actions and/or status checks UI. In addition, it will simplify the future changes if/when we decide to move functions support to its own builder tag.

GUS-W-14121701.


Malax (approved):
👍🏻 - we need to update the required checks for this repo before merging.

PR #392, opened 2023-09-25
Add a Salesforce Functions builder variant

edmorley:
We want to move Salesforce Functions buildpacks to their own builder image variant to more cleanly separate the two use cases (particularly given that Salesforce Functions is now deprecated), and prevent a number of issues (such as builds on the functions build system being able to pass detection even when the match against a non-functions buildpack).

For more details, see:
https://salesforce.quip.com/MoDFAX3MEf8b

The first stage of this is to create the new builder image variant.

This new variant is only being published to the internal private registry, since it's only being used by Kodon (the SF CLI no longer supports using containers to run functions) so doesn't need to be available on Docker Hub.

Once it's released, we can then switch Kodon to using the new tag, deploy Kodon to all functions spaces, and then finally remove the functions buildpacks from `heroku/builder:22`.

We'll also need to adjust the CNB release automation so that it updates the new image variant too:
https://github.com/heroku/languages-github-actions/blob/8f1d3319996df5ec9eb3b4c9c7c0d5de78fa9dba/.github/workflows/_buildpacks-release.yml#L315

GUS-W-14179967.

### DIFF
--- a/.github/workflows/build-test-publish.yml
+++ b/.github/workflows/build-test-publish.yml
@@ -25,7 +25,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        builder: ["buildpacks-20", "builder-classic-22", "builder-22"]
+        builder: ["buildpacks-20", "builder-classic-22", "builder-22", "salesforce-functions"]
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


.github/workflows/build-test-publish.yml
```diff
@@ -34,7 +34,11 @@ jobs:
       - name: Export Docker images from the Docker daemon
         env:
           BUILDER: ${{ matrix.builder }}
-        run: docker save ${{ matrix.builder }} heroku/heroku:${BUILDER##*-}-cnb | zstd -T0 --long=31 -o images.tar.zst
```
edmorley, line 37: This previous implementation for calculating the run image tag assumed the run image tag name used a naming pattern related to the directory name in this repository. 

Since the new builder image variant does not (the subdirectory is `salesforce-functions`, this has now had to be switched to looking up the run image from the builder.toml file itself - which is really a more correct solution anyway.

edmorley:
Note to self: When this is ready to be merged, also adjust the required status checks in the branch protection settings.

Malax (approved):

edmorley:
Branch protection list of required CI checks updated.

PR #394, opened 2023-09-26
Add `heroku/builder:20`

edmorley:
The builder images in this repository fall into two main groups:
1. The legacy `heroku/buildpacks:*` builder namespace that use a mixture of native CNBs and shimmed classic buildpacks.
2. The newer `heroku/builder:*` builder namespace that uses only native (non-shimmed) CNBs.

Previously, if someone wanted to use the newer builder image type they had to use `heroku/builder:22` which is based on the Heroku-22 stack (Ubuntu 22.04). 

Now, if someone needs to use Heroku-20  (Ubuntu 20.04), they can do so via a new `heroku/builder:20` image, rather than needing to fall back to the older `heroku/buildpacks:20` image and its shimmed classic buildpacks.

This will make it easier for us to deprecate `heroku/buildpacks:20` in future PRs, since there will now be an easier upgrade paths for users that need eg older language runtime versions of Python/Ruby/PHP etc.

The new `heroku/builder:20` builder image variant is identical to `heroku/builder:22`, except:
- it uses Heroku-20 based run/build-images (instead of Heroku-22 based)
- it doesn't include the java/node.js functions buildpacks (which are due to be removed from `heroku/builder:22` soon in favour of the newly added internal Salesforce Functions image variant).

When this is merged, we'll also need to:
- add the new CI jobs to the branch protection required CI jobs list
- adjust the CNB release automation so that it updates this new builder too (heroku/languages-github-actions#144)

For more details, see:
https://salesforce.quip.com/0JtbAYiWZYk6

GUS-W-14184901.

### DIFF
--- a/.github/workflows/build-test-publish.yml
+++ b/.github/workflows/build-test-publish.yml
@@ -25,7 +25,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        builder: ["buildpacks-20", "builder-classic-22", "builder-22", "salesforce-functions"]
+        builder: ["buildpacks-20", "builder-classic-22", "builder-20", "builder-22", "salesforce-functions"]
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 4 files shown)


dzuelke (approved):

edmorley:
Branch protection list of required CI checks updated.

edmorley:
🎉  https://hub.docker.com/r/heroku/builder/tags?page=1&name=20

PR #404, opened 2023-10-19
Add automation for updating builder lifecycle version

edmorley:
Previously we had to manually keep track of [upstream lifecycle releases](https://github.com/buildpacks/lifecycle/releases), and then open PRs to update the builder manifest for each builder with the new version. 

For example:
https://github.com/heroku/cnb-builder-images/pull/389

Now, every Monday at 8am UTC, a GitHub Actions job will check for updates, and open a PR if the latest lifecycle version is newer than that used by the `heroku/builder:22` builder. (This means lifecycle version pinning will be supported too for eg our legacy images, since only builder images on the latest version will end up being updated each time.)

This new workflow can also be triggered manually thanks to the `workflow_dispatch` annotation.

GUS-W-14329446.

### DIFF
--- a/.github/workflows/update-lifecycle.yml
+++ b/.github/workflows/update-lifecycle.yml
@@ -1,0 +1,71 @@
+name: Update lifecycle
+
+on:
+  schedule:
+    # Every Monday at 8am UTC.
+    - cron: '0 8 * * MON'
+  workflow_dispatch:
+
+jobs:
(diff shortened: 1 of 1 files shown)


Malax:
This is super neat! :) Thanks for adding this one! 🎉

Malax (approved):

PR #418, opened 2023-10-25
Lockfiles for Node Functions examples

colincasey:
To unblock #417.  There is no reason why these examples shouldn't have lockfiles in place.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .DS_Store
 .idea
 *.iml
+node_modules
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
 .DS_Store
 .idea
 *.iml
+node_modules
```
edmorley, line 4: (Missing end of file newline)

```suggestion
node_modules

```

edmorley (approved):

PR #336, opened 2023-05-09
Make Heroku-18 builds fail with a skippable EOL warning

edmorley:
The Heroku-18 stack reached end-of-life on April 30th 2023, and as of May 1st 2023, builds are no longer possible on the (default) non-CNB Heroku build system, and the underlying stack images will no longer
 receive security updates (beyond whatever updates Canonical choose to
release in May). 

However, anyone using the `heroku/buildpacks:18` CNB builder image may not realise that it's now EOL and soon to be insecure. They may also not realise that the builder image has moved from the `heroku/buildpacks` Docker repo to `heroku/builder`.

In order to raise awareness of both of these, an EOL buildpack has been added, which fails the build with an EOL warning, unless the env var `ALLOW_INSECURE_HEROKU_18_BUILDER` is set to `1` during the build.

This should affect very few users, since Heroku is not yet using CNBs by default, and Salesforce Functions use the `heroku/builder:22` builder image instead.

This buildpack was written in bash, since:
- the buildpack functionality we need is extremely simple (checking an env var, printing a message and exiting 1 or 0)
- the buildpack (and other Heroku-18 related files) are going to be deleted in a few weeks anyway, so its not worth spending time getting a Rust build workflow working with the existing complex GitHub Actions workflow here

GUS-W-13143190.

PR #429, opened 2023-10-27
Add deprecation warnings to legacy builder images

edmorley:
Adds deprecation warnings to the `heroku/builder-classic:22` and `heroku/buildpacks:20` CNB builder images, to raise awareness that they are no longer supported/recommended.

The warnings have been added using a warnings buildpack, similar to the approach used for the Heroku-18 EOL in:
https://github.com/heroku/cnb-builder-images/pull/336

For now these are just warnings, however, in the future will be changed to an error (skippable via env var, like in the PR linked above).

Since these are warnings, the buildpack has been added to the end of each buildpack order group, such that the message is at the end of the overall build log, and so hopefully more visible to end users. (Once the warning is turned into an error, the buildpack should be moved to the start of the order groups, so the build fails early for improved UX.)

A separate mostly-copy-pasted buildpack was used for each builder, since there is no other easy way to customise the message shown for each (and I wanted to include the name of the deprecated builder in the warning message, so users know what image name to grep for).

These buildpacks were written in bash (rather than in Rust, using [libcnb.rs](https://github.com/heroku/libcnb.rs)), since:
- the buildpack functionality we need is extremely simple
- the buildpack is temporary (it will be deleted when we stop updating these builders in the future), so we don't need/want long term published images on CNB registry or to have to set up a Rust compilation/packaging step in this repo.

The buildpacks give zero lint warnings when checked with [shellcheck](https://github.com/koalaman/shellcheck) locally.

GUS-W-14194729.
GUS-W-14194736.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,10 +10,10 @@ The builder images use Heroku's [stack images](https://github.com/heroku/stack-i
 | Builder Image                                       | Base Image                                  | Status      |
 |-----------------------------------------------------|---------------------------------------------|-------------|
 | [`heroku/buildpacks:18`][buildpacks-tags]           | [`heroku/heroku:18-cnb-build`][heroku-tags] | End-of-life |
-| [`heroku/buildpacks:20`][buildpacks-tags]           | [`heroku/heroku:20-cnb-build`][heroku-tags] | Available   |
+| [`heroku/buildpacks:20`][buildpacks-tags]           | [`heroku/heroku:20-cnb-build`][heroku-tags] | Deprecated  |
 | [`heroku/builder:20`][builder-tags]                 | [`heroku/heroku:20-cnb-build`][heroku-tags] | Available   |
 | [`heroku/builder:22`][builder-tags]                 | [`heroku/heroku:22-cnb-build`][heroku-tags] | Recommended |
-| [`heroku/builder-classic:22`][builder-classic-tags] | [`heroku/heroku:22-cnb-build`][heroku-tags] | Available   |
+| [`heroku/builder-classic:22`][builder-classic-tags] | [`heroku/heroku:22-cnb-build`][heroku-tags] | Deprecated  |
 
 [`heroku/builder`][builder-tags] builder images feature Heroku's native Cloud Native Buildpacks. These buildpacks are optimized and make use of many CNB features. These builder images support Go, Java (Maven, Gradle), Node.js, PHP, Python, Ruby, Scala and Typescript codebases.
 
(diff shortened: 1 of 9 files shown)


edmorley:
Example warning for `heroku/builder-classic:22`:
https://github.com/heroku/cnb-builder-images/actions/runs/6669195890/job/18126518745?pr=429#step:6:248

And for `heroku/buildpacks:20`:
https://github.com/heroku/cnb-builder-images/actions/runs/6669195890/job/18126512300?pr=429#step:6:241

And locally in a console that supports colour:

Console screenshot

dzuelke:
I am not the biggest fan of the `echo_stderr` name, since it doesn't only redirect to stderr, but also colorize the text.

About the HEREDOCs...

Instead of this workaround:

```bash
read -r -d '' EOL_MESSAGE <<-'EOF' || true
…
EOF

echo_stderr "${EOL_MESSAGE}"
```

You can simplify it to this:

```bash
echo_stderr "$(cat <<'EOF'
…
EOF
)"
```

Alternatively, in the helper function, you can call `cat` without arguments instead of echoing `$1`, like so:

```bash
function echo_stderr() {
  # send all of our output to stderr
  exec 1>&2
  # if arguments are given, redirect them to stdin
  # this allows the funtion to be invoked with a string argument, or with stdin, e.g. via <<-EOF
  (( $# )) && exec <<< "$@"
  local ansi_red="\033[1;31m"
  local ansi_reset="\033[0m"
  echo -ne "\n${ansi_red}"
  cat
  echo -e "${ansi_reset}\n"
}
```

The invocation then becomes much cleaner:

```bash
echo_stderr <<'EOF'
…
EOF
```

Due to the `(( $# )) && exec <<< "$@"` line, you can still pass simple string arguments.

edmorley:
> You can simplify it to this:

@dzuelke The reason for the current implementation, is that once it's turned into an "error if env var not set" the message needs to be used twice, so needs to be assigned to a variable.

This can be seen in the PR linked in the PR description (and why I mentioned in the description that this would be turned into an error in the future):
https://github.com/heroku/cnb-builder-images/pull/336

edmorley:
> I am not the biggest fan of the `echo_stderr` name

I've renamed it to `display_error`.

Malax (approved):

edmorley:
I've also since found:
https://gitlab.com/gitlab-org/cluster-integration/auto-build-image/-/issues/73
https://gitlab.com/gitlab-org/cluster-integration/auto-build-image/-/merge_requests/159

edmorley:
I plan on upgrading these deprecation warnings to errors (that are skippable with an env var) in the next couple of weeks.

PR #434, opened 2023-11-10
Heroku 22 Stack Image Hierarchy diagram

ryanbrainard:
As someone new to CNB, the relationship between all the various images and how they relate to the Heroku stack image and the CNB concepts is pretty confusing. To help wrap my head around it all, I ended up making a diagram. Sharing here in case there is interest in including it in the documentation.


dzuelke:
Cool, we need more illustrations like this again in our docs I think.

Would you mind turning this into an [inline](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-diagrams#creating-mermaid-diagrams) [Mermaid](http://mermaid.js.org/) diagram?

The coloring etc might then be not as nice, but it will be so, so much easier to keep this diagram up to date going forward - we'll likely make some changes to the hierarchy in the next few months already, for which we'll need the source files (or re-do the diagram in some app).

Keeps repo size small, too ;)

ryanbrainard:
@dzuelke I opened https://github.com/heroku/cnb-builder-images/pull/437 with the same thing in Mermaid. (I did it on a separate branch so we can delete this branch to avoid the image from becoming part of the git history.)

PR #440, opened 2023-11-23
Improve documentation and description metadata

edmorley:
The repo README has been overhauled, to:
- Give more context as to what a builder is, and link to the upstream docs.
- Emphasise that these images are experimental and not yet used for "normal" Heroku builds.
- Make it easier to understand the difference between the image variants, including attributes such as run image and lifecycle version.
- Make the usage instructions more accessible for beginners.
- Give users breadcrumbs to follow when trying to work out where to file issues.

The "update lifecycle" GitHub Actions workflow has been modified so that it now also modifies the lifecycle version in the README in addition to that in the builder manifests.

Lastly, the description metadata used for each builder (that is shown by `pack builder inspect`) has also been updated to mention deprecations and bring the descriptions inline with that used upstream:
https://github.com/buildpacks/pack/blob/9c8223caef09a8f73b9c145c0ccf03ab704aac53/internal/builder/suggested_builder.go#L16-L23

GUS-W-14194789.

### DIFF
--- a/.github/workflows/update-lifecycle.yml
+++ b/.github/workflows/update-lifecycle.yml
@@ -42,6 +42,9 @@ jobs:
         # that any legacy builder images pinned to older lifecycle versions are not updated too.
         run: sed --in-place --expression 's/^version = "${{ steps.existing-version.outputs.version }}"$/version = "${{ steps.latest-version.outputs.version }}"/' */builder.toml
 
+      - name: Update lifecycle versions in README to match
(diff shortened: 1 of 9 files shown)


edmorley:
Rendered:
* Before: https://github.com/heroku/cnb-builder-images/blob/main/README.md
* After: https://github.com/heroku/cnb-builder-images/blob/edmorley/overhaul-docs/README.md

dzuelke (approved):

colincasey (approved):

PR #463, opened 2024-02-05
Use `--force-color` with `pack build` invocations

edmorley:
As of Pack 0.33.0 (which this repo is using thanks to #462), there is a new `--force-color` argument that makes `pack build` output logs in colour even if a TTY wasn't detected (as is the case on GHA):
https://github.com/buildpacks/pack/releases/tag/v0.33.0

This will allow us to see the colour output from the new build output implementation.

GUS-W-14968745.


edmorley:
Before:
https://github.com/heroku/cnb-builder-images/actions/runs/7782352648/job/21218661625#step:6:1

After:
https://github.com/heroku/cnb-builder-images/actions/runs/7783873888/job/21223304231#step:6:1

colincasey (approved):

PR #476, opened 2024-03-08
Fix broken upstream docs URL in deprecation message

edmorley:
The upstream CNB project has rearranged a number of the docs pages, breaking some existing links such as:
https://buildpacks.io/docs/tools/pack/cli/pack_config_default-builder/

I've filed https://github.com/buildpacks/docs/issues/673 for potentially adding redirects, however, for the deprecation warning use-case, it makes more sense for us to just inline the CLI command rather than link out to the docs.


Malax (approved):

edmorley:
Hmm so trying now the link works (it didn't last night). It seems this may have been fixed by:
https://github.com/buildpacks/docs/pull/671

I think it still makes sense for us to inline the CLI command rather than link to an external page, so I'll merge this anyway :-)

PR #478, opened 2024-03-08
Upgrade legacy builder deprecation warnings to errors

### LINKED ISSUE #474: Upgrade legacy builder image deprecation warnings to an error
In June 2022, as part of the Heroku-22 release, we switched to a new Docker Hub repo for our builder images (from `heroku/buildpacks:*` to `heroku/builder:*`).

In addition to the change in repository name, the new `heroku/builder:*` namespaced builder images contain only Heroku's next-gen (fully native) Cloud Native Buildpacks, whereas the old `heroku/buildpacks:*` builder images contain a mixture of [shimmed](https://github.com/heroku/cnb-shim) and native CNBs (since back when that builder image was created, we didn't yet have native CNBs for every supported language, so some had to be substituted using shimmed classic Heroku buildpacks).

For more details on the differences between the images, see:
https://github.com/heroku/cnb-builder-images/blob/main/README.md#available-images

To raise awareness of the deprecated nature of the old images, in October 2023:
- Deprecation warnings for `heroku/buildpacks:20` were added to build logs in #429. We also deprecated `heroku/builder-classic:22` at the same time, since it also uses shimmed CNBs.
- Outreach was performed to various popular OSS projects found using the deprecated images: https://github.com/heroku/cnb-builder-images/pull/429#issuecomment-1785095460

IMO it's now time to upgrade this deprecation warnings to errors, since:
- We're about to add many more image variants (due to multi-architecture + Ubuntu 24.04 support), so it would be helpful to reduce the number of supported builder images prior to that.
- We're having to jump through additional hoops for these older images, like pinning the `lifecycle` version due to `cnb-shim` not being compatible with newer Buildpack API versions (see [here](https://github.com/heroku/cnb-builder-images/pull/431#issuecomment-1809968042)).
- It has been 21 months since we introduced the new `heroku/builder:*` repo and coming up to six months of us outputting loud deprecation warnings.
- All of our builder images are still experimental/pre-GA/not used in production at Heroku (and marked as such in the README), so the amount of notice we've given is IMO already more than reasonable (once GA there will be more stability and longer deprecation periods).
- We're preparing for a more official beta phase for our CNBs, so would like to reduce the chance of users trying out a deprecated image by mistake / to clean things up in general prior to then.

At least initially, we can make these errors skippable by setting an env var, similar to what we did for Heroku-18's EOL in #336.

I think we should do this within the next couple of weeks.

cc @heroku/languages

edmorley:
This upgrades the deprecation warnings added in #429 to errors, for the reasons mentioned in #474. It uses an approach similar to that used for the Heroku-18 EOL in #336.

These errors can be skipped by setting the env var: `ALLOW_EOL_SHIMMED_BUILDER=1`

For example:

`pack build --env ALLOW_EOL_SHIMMED_BUILDER=1 ...`

However, users are strongly encouraged to migrate from the legacy shimmed `heroku/buildpacks:20` and `heroku/builder-classic:22` images to the new `heroku/builder:*` builder images.

For the differences between the available images, see:
https://github.com/heroku/cnb-builder-images#available-images

The buildpack has also been moved to the start of each `[[order]]` group (rather than the end), so that the build fails early, rather than wasting the user/build system's time with a build that's only going to fail. (We had the buildpack at the end of the group before, to make the EOL warning more visible, since often users won't scroll back through the logs, and only see what's printed at the end of the build log in their console.)

Example error message:

error-message

See also:
https://github.com/heroku/cnb-builder-images/actions/runs/8204908398/job/22440626428?pr=478#step:6:30

Closes #474.
GUS-W-15212520.

### DIFF
--- a/.github/workflows/build-test-publish.yml
+++ b/.github/workflows/build-test-publish.yml
@@ -82,7 +82,7 @@ jobs:
       - name: Load Docker images into the Docker daemon
         run: zstd -dc --long=31 images.tar.zst | docker load
       - name: Build getting started guide image
-        run: pack build getting-started --force-color --builder ${{ matrix.builder }} --trust-builder --pull-policy never
+        run: pack build getting-started --force-color --builder ${{ matrix.builder }} --trust-builder --pull-policy never --env ALLOW_EOL_SHIMMED_BUILDER=1
       - name: Start getting started guide image
         # The `DYNO` env var is set to more accurately reflect the Heroku environment, since some of the getting
         # started guides use the presence of `DYNO` to determine whether to enable production mode or not.
--- a/README.md
+++ b/README.md
@@ -17,15 +17,15 @@ For more information, see: [What is a builder?](https://buildpacks.io/docs/conce
 ## Available images
 
 > [!WARNING]
-> The `heroku/buildpacks:*` and `heroku/builder-classic:*` builder image variants are deprecated,
+> The `heroku/buildpacks:*` and `heroku/builder-classic:*` builder image variants have been sunset,
 > since they use classic Heroku buildpacks shimmed for compatibility with the CNB specification,
 > rather than Heroku's next-generation Cloud Native Buildpacks.
 
 | Builder Image                                       | Base Build Image                            | Base Run Image                        | Lifecycle Version | Buildpack Types  | Status      |
 |-----------------------------------------------------|---------------------------------------------|---------------------------------------|-------------------|------------------|-------------|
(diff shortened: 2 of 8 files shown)


edmorley:
I've opened this early as a draft to make it easier for anyone following #474 to see what's upcoming, however, I won't be merging this PR until slightly later this month.

edmorley:
I plan to merge this at the start of next week.

Malax (approved):

PR #487, opened 2024-03-16
Update readme

runesoerensen:


README.md
```diff
@@ -5,14 +5,13 @@
 > [!IMPORTANT]
 > These images are experimental and still in active development: [heroku/roadmap#20](https://github.com/heroku/roadmap/issues/20)
```
edmorley, line 6: Do we need to adjust the wording here too? I can't decide :-)
(The wording is technically still correct, but it's the only place we're using the phrasing "experimental".)

edmorley: It was me who added the "experimental" wording a while back - so feel free to replace with "preview" :-)

edmorley (approved):

PR #504, opened 2024-05-02
Introduce builder-24 with multi-arch support

joshwlewis:
This introduces a multi-arch builder with buildpacks that support both `linux/amd64` and `linux/arm64` based on heroku-24. For now, that's limited to `heroku/go` and `heroku/procfile`. More languages will be added as they gain `linux/arm64` support.

This makes use of the GitHub Actions arm runners beta to build the arm images. These runners don't have `docker` so there are some additional manual install steps.

Example successful run: https://github.com/heroku/cnb-builder-images/actions/runs/8941039965/job/24560480143

GUS-W-14658981.
GUS-W-15158337.


.github/workflows/build-test-publish.yml
```diff
@@ -142,8 +174,8 @@ jobs:
             exit 1
           fi
 
-  publish:
-    runs-on: ubuntu-22.04
+  publish-images:
+    runs-on: ${{ matrix.arch == 'arm64' && 'pub-hk-ubuntu-22.04-arm-large' || 'ubuntu-22.04' }}
```
edmorley, line 178: The `publish-images` step is only docker push/tagging existing images. As such, it seems it can run on any runner, and so the `runs-on` property here could be `ubuntu-22.04` for all permutations in the matrix (including the arch-arm64 jobs)? This would save having to do the manual Docker install, plus mean being able to use the warm `ubuntu-22.04` runner pool GitHub has, instead of the slower org ~just-in-time-provisioned instances.

README.md
```diff
@@ -20,13 +20,14 @@ For more information, see: [What is a builder?](https://buildpacks.io/docs/for-p
...
+| [`heroku/builder-classic:22`][builder-classic-tags] | [`heroku/heroku:22-cnb-build`][heroku-tags] | [`heroku/heroku:22-cnb`][heroku-tags] | 0.17.6            | Shimmed          | End-of-life    |
+| [`heroku/builder:20`][builder-tags]                 | [`heroku/heroku:20-cnb-build`][heroku-tags] | [`heroku/heroku:20-cnb`][heroku-tags] | 0.19.3            | Native           | Available      |
+| [`heroku/builder:22`][builder-tags]                 | [`heroku/heroku:22-cnb-build`][heroku-tags] | [`heroku/heroku:22-cnb`][heroku-tags] | 0.19.3            | Native           | Recommended    |
+| [`heroku/builder:24`][builder-tags]                 | [`heroku/heroku:24-build`][heroku-tags]     | [`heroku/heroku:24`][heroku-tags]     | 0.19.3            | Native           | In Development |
 
 The builder images above include buildpack support for the following languages: Go, Java, Node.js, PHP, Python, Ruby & Scala. The `heroku/builder-classic:22` builder image variant additionally supports Clojure.
```
edmorley, line 32: Do we need to mention ARM64 support somewhere in the README? Sadly the table is already pretty crowded, so maybe it would have to be a mention somewhere in the paragraphs after the table?

edmorley (approved):
🎉

.github/workflows/build-test-publish.yml
joshwlewis: This step is also loading images into the daemon and tagging images in the daemon. Linux docker's daemon doesn't handle storage/retrieval of cross-architecture images well (at all?).

If this step were retagging images already in a remote registry OR we were using Docker Desktop, we wouldn't need to match architectures.

README.md
joshwlewis: Yeah, maybe so. I'll add a paragraph. Should we also mention it doesn't support all languages there too?

.github/workflows/build-test-publish.yml
edmorley: Ah right, sounds like another thing that works fine on Docker Desktop on macOS (even without the containerd snapshotter enabled), but that has rough edges with the Linux docker daemon. In which case, ignore this then :-)

README.md
edmorley: I guess another option might be to skip adding anything to the README for now (like we did for the base-images repo), and only add an entry once we have more languages added?

I was also going to remove the publishing of the legacy builders shortly, and clean up the readme and table at the same time (eg remove the native vs shimmed column), which might free up some table space for a "Supported Architectures" column etc?

edmorley: I'm fine with any of the above tbh (incl just leaving as it is in the current version of the PR) - whatever you prefer :-)

joshwlewis: Yeah, I guess I'll leave it as-is for now. I look forward to the cleanup where we can remove several other sources of confusion.

PR #514, opened 2024-05-03
Stop publishing `heroku/buildpacks:20` and `heroku/builder-classic:22`

### LINKED ISSUE #512: Stop publishing updates for the legacy builder images
The `heroku/buildpacks:20` and `heroku/builder-classic:22` builder images were:
- superseded by the `heroku/builder:*` images in June 2022
- officially deprecated in Oct 2023 (#429)
- made to error by default in March 2024 (#474)

It's now time to stop updating these images.

This will help offset the growth in our CI/publishing matrix after the recent addition of `heroku/builder:24` (which added many more jobs, given it's also multi-arch).

Before we stop publishing the images, we'll want to update the EOL messaging in them one last time, to make the message about "stopping receiving updates in the future" be current/past tense instead.

GUS-W-15677907.

edmorley:
Since these images were:
- superseded by the `heroku/builder:*` images in June 2022
- officially deprecated in Oct 2023 (#429)
- made to error by default in March 2024 (#474)

Builds can still continue using the last version of these builders published to Docker Hub, and the resultant app images can still be rebased onto newer run images if needed.

I've also cleaned up the available images table in the README, to make it display a little better. (There was too much content before, causing the table to wrap awkwardly when viewed in the narrow GitHub repo-root view.)

Closes #512.
GUS-W-15677907.


Makefile
```diff
@@ -3,15 +3,7 @@
...
 build:
-	@pack builder create heroku/buildpacks:20 --config buildpacks-20/builder.toml --pull-policy always
 	@pack builder create heroku/builder:20 --config builder-20/builder.toml --pull-policy always
 	@pack builder create heroku/builder:22 --config builder-22/builder.toml --pull-policy always
-	@pack builder create heroku/builder-classic:22 --config builder-classic-22/builder.toml --pull-policy always
+	@pack builder create heroku/builder:24 --config builder-24/builder.toml --pull-policy always
 	@pack builder create salesforce-functions --config salesforce-functions/builder.toml --pull-policy always
-
-# The salesforce-functions image isn't published to Docker Hub, so is intentionally absent here.
-publish: build
```
edmorley, line 13: I've removed this publish Makefile target since we don't (and shouldn't) be publishing locally - particularly now there are many more details to take into account given multi-arch (which means even if the GHA pipeline went down, this current Makefile target wouldn't be sufficient to roll out a release anyway).

edmorley:
Before:
https://github.com/heroku/cnb-builder-images/blob/d74b3df909399e28e5ab39fad894ee86d3830a83/README.md

After:
https://github.com/heroku/cnb-builder-images/blob/c3f980b53ee913661a4325328735972660839b0a/README.md

edmorley:
I've also opened https://github.com/heroku/languages-github-actions/pull/220 so the CNB release automation doesn't try to update manifest files that will no longer exist in this repo once this PR merges.

runesoerensen (approved):

PR #522, opened 2024-05-20
Cleanup workflow job definitions

edmorley:
Now that we build/test ARM64 images too, the job names shown in the GitHub Actions UI have become quite hard to read.

This adjust the job names and matrix parameter order (which determines the order in which parameters appear in the auto-generated job names) to make them easier to read in the UI.

In addition, the Heroku-24 test entries have been changed to use `exclude` rather than `include` to reduce the amount of boilerplate, and also mean the jobs display in the correct sequence in the UI.

### DIFF
--- a/.github/workflows/build-test-publish.yml
+++ b/.github/workflows/build-test-publish.yml
@@ -21,15 +21,21 @@ defaults:
 jobs:
 
   create:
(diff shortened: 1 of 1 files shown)


edmorley:
## Before

before

## After

after

schneems (approved):

PR #529, opened 2024-05-30
Use pub-hk-ubuntu-22.04-arm-medium

dzuelke:
Smaller runners are more plentiful, and we no longer need to install Docker, either!


.github/workflows/build-test-publish.yml
```diff
@@ -93,14 +85,6 @@ jobs:
       - name: Checkout
```
edmorley, line 85: This checkout step can be removed too (it was only needed so that this specific job could access/run the `bin/setup_docker_ci` script), plus the same checkout step in the `publish` jobs, plus also the `bin/setup_docker_ci` script itself.

joshwlewis: Fixed in https://github.com/heroku/cnb-builder-images/pull/529/commits/4064944309043bb95f5e33059fe46367269b526a.

joshwlewis (approved):

PR #557, opened 2024-07-25
Copy builder images to public ecr

joshwlewis:
This PR changes the publish automation to additionally push builder images to `public.ecr.aws/heroku/builder`. This aims to alleviate some Docker Hub rate limit concerns and enables faster downloads in some cloud environments. The primary and canonical location for these images is still `heroku/builder` (on dockerhub).

This automation makes use of GitHub OIDC to authenticate with AWS so no IAM users or access tokens are in use.

[GUS](https://gus.lightning.force.com/a07EE00001qVbCRYA0)

### DIFF
--- a/.github/workflows/build-test-publish.yml
+++ b/.github/workflows/build-test-publish.yml
@@ -235,7 +235,7 @@ jobs:
       - name: Tag builder and push to public ECR
         if: matrix.tag_ecr_public != ''
         run: |
-          ECR_PUBLIC_IMAGE_URI='public.ecr.ams/${{ matrix.tag_ecr_public }}'
+          ECR_PUBLIC_IMAGE_URI='public.ecr.aws/${{ matrix.tag_ecr_public }}'
           set -x
           docker tag '${{ matrix.builder }}' "${ECR_PUBLIC_IMAGE_URI}"
           docker push "${ECR_PUBLIC_IMAGE_URI}"
--- a/.github/workflows/build-test-publish.yml
+++ b/.github/workflows/build-test-publish.yml
@@ -165,19 +165,22 @@ jobs:
         include:
           - builder: builder-20
             arch: amd64
-            tag_public: heroku/builder:20
+            tag_docker_hub: heroku/builder:20
           - builder: builder-22
             arch: amd64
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test-publish.yml
```diff
@@ -157,27 +158,30 @@ jobs:
 
   publish-image:
     runs-on: ${{ matrix.arch == 'arm64' && 'pub-hk-ubuntu-24.04-arm-medium' || 'ubuntu-24.04' }}
-    if: success() && github.ref == 'refs/heads/main'
+    if: success() # && github.ref == 'refs/heads/main'
```
edmorley, line 161:
```suggestion
    if: success() && github.ref == 'refs/heads/main'
```

.github/workflows/build-test-publish.yml
```diff
@@ -202,35 +206,66 @@ jobs:
...
+      - name: Tag builder and push to public ECR
+        if: matrix.tag_ecr_public != ''
+        run: |
+          ECR_PUBLIC_IMAGE_URI='public.ecr.aws/${{ matrix.tag_ecr_public }}'
+          set -x
+          docker tag '${{ matrix.builder }}' "${ECR_PUBLIC_IMAGE_URI}"
+          docker push "${ECR_PUBLIC_IMAGE_URI}"
 
-  publish-manifest:
+  publish-manifests:
```
edmorley, line 243: For the other job names we're using the singular even if there is a matrix, since each job only does a single action:

```
create (builder-24, arm64)
test (go, builder-24, arm64)
publish-image (builder-20, amd64, heroku/builder:20)
```

ie: The `publish-image (builder-20, amd64, heroku/builder:20)` job only publishes one local image.

```suggestion
  publish-manifest:
```

edmorley (approved):
Thank you! I love the use of OIDC here over IAM creds 😄 

I will separately review the IAM role in our AWS account, if only to learn more about how that works.

.github/workflows/build-test-publish.yml
joshwlewis: Ah yeah, I had a previous iteration where it was pushing multiple per job.

PR #571, opened 2024-08-29
Add .NET buildpack

runesoerensen:
Adding the .NET buildpack to `heroku/builder:22` and `heroku/builder:24`.

### DIFF
--- a/.github/workflows/build-test-publish.yml
+++ b/.github/workflows/build-test-publish.yml
@@ -79,8 +79,6 @@ jobs:
             arch: arm64
           - builder: builder-20
             language: dotnet
-          - builder: builder-22
-            language: dotnet
 
     steps:
       - name: Checkout getting started guide
--- a/.github/workflows/build-test-publish.yml
+++ b/.github/workflows/build-test-publish.yml
@@ -69,7 +69,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        language: ["go", "gradle", "java", "node-js", "php", "python", "ruby", "scala"]
+        language: ["dotnet", "go", "gradle", "java", "node-js", "php", "python", "ruby", "scala"]
         builder: ["builder-20", "builder-22", "builder-24"]
         arch: ["amd64", "arm64"]
         exclude:
@@ -77,6 +77,11 @@ jobs:
             arch: arm64
           - builder: builder-22
             arch: arm64
+          - builder: builder-20
+            language: dotnet
+          - builder: builder-22
+            language: dotnet
+          
 
     steps:
       - name: Checkout getting started guide
--- a/.github/workflows/build-test-publish.yml
+++ b/.github/workflows/build-test-publish.yml
@@ -81,7 +81,6 @@ jobs:
             language: dotnet
           - builder: builder-22
             language: dotnet
-          
 
     steps:
       - name: Checkout getting started guide
--- a/README.md
+++ b/README.md
@@ -16,11 +16,14 @@ in each base image, see [this Dev Center article](https://devcenter.heroku.com/a
 
 ## Available images
 
-| Builder Image                     | OS           | Architectures | Supported Language Buildpacks                      | Status      |
-|-----------------------------------|--------------|---------------|----------------------------------------------------|-------------|
-| [heroku/builder:20][builder-tags] | Ubuntu 20.04 | AMD64         | Go, Java, Node.js, PHP, Python, Ruby & Scala       | Deprecated  |
-| [heroku/builder:22][builder-tags] | Ubuntu 22.04 | AMD64         | .NET, Go, Java, Node.js, PHP, Python, Ruby & Scala | Available   |
-| [heroku/builder:24][builder-tags] | Ubuntu 24.04 | AMD64 + ARM64 | .NET, Go, Java, Node.js, PHP, Python, Ruby & Scala | Recommended |
+| Builder Image                     | OS           | Supported Architectures | Default Run Image                   | Lifecycle Version | Status      |
+|-----------------------------------|--------------|-------------------------|-------------------------------------|-------------------|-------------|
+| [heroku/builder:20][builder-tags] | Ubuntu 20.04 | AMD64                   | [heroku/heroku:20-cnb][heroku-tags] | 0.20.1            | Deprecated  |
+| [heroku/builder:22][builder-tags] | Ubuntu 22.04 | AMD64                   | [heroku/heroku:22-cnb][heroku-tags] | 0.20.1            | Available   |
+| [heroku/builder:24][builder-tags] | Ubuntu 24.04 | AMD64 + ARM64           | [heroku/heroku:24][heroku-tags]     | 0.20.1            | Recommended |
+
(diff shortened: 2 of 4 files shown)


.github/workflows/build-test-publish.yml
```diff
@@ -69,14 +69,19 @@ jobs:
...
         exclude:
           - builder: builder-20
             arch: arm64
           - builder: builder-22
             arch: arm64
+          - builder: builder-20
+            language: dotnet
+          - builder: builder-22
+            language: dotnet
+
```
edmorley, line 84:
```suggestion
```

README.md
```diff
@@ -22,7 +22,8 @@ in each base image, see [this Dev Center article](https://devcenter.heroku.com/a
 | [heroku/builder:22][builder-tags] | Ubuntu 22.04 | AMD64                   | [heroku/heroku:22-cnb][heroku-tags] | 0.20.1            | Available   |
 | [heroku/builder:24][builder-tags] | Ubuntu 24.04 | AMD64 + ARM64           | [heroku/heroku:24][heroku-tags]     | 0.20.1            | Recommended |
 
-The builder images above include buildpack support for the following languages: Go, Java, Node.js, PHP, Python, Ruby & Scala.
+The builder images above all include buildpack support for the following languages: Go, Java, Node.js, PHP, Python, Ruby & Scala.
+Additionally, `heroku/builder:24` include buildpack support for .NET applications.
```
edmorley, line 26:
```suggestion
Additionally, `heroku/builder:24` includes buildpack support for .NET applications.
```

edmorley, line 26: I'm presuming the dotnet buildpack would actually work on the `heroku/builder:{20,22}` builders, given that the runtime isn't distro-specific and so the buildpack's declared `[[targets]]` doesn't limit by distro?

ie: If someone manually ran `pack build --builder heroku/builder:22 --buildpack heroku/dotnet --buildpack heroku/procfile` (or had the equivalent in their `project.toml`), then I presume things would work? That is, us excluding it from older builder images won't actually stop people using it if they want to.

If so, could/should we include the dotnet buildpack in the older builder images too, if only so we don't have to say "this image has a different list of languages"? 

(I can see pros/cons both ways; I'm just curious)

runesoerensen: Yes it works just fine on the `heroku/builder:{20,22}` builders. And people can use it whether we include it in our older builders or not. I haven't "manually" tested those as vigorously, but I know for sure that the [dotnet-getting-started](https://github.com/heroku/buildpacks-dotnet) work on `builder:22` and `builder:20`.

> If so, could/should we include the dotnet buildpack in the older builder images too, if only so we don't have to say "this image has a different list of languages"?

I had the same thought while making that change, and was going to ask for feedback on how the phrasing could be improved (but you obviously beat me to it) :) I'd personally also be inclined to include it in the earlier builders, if only to avoid having to write .NET specific language in our builder docs.

On the other hand, adding it to`builder:20` might be a bit odd seeing as it's deprecated?

colincasey: I don't think we should add it to the deprecated `builder:20` but adding it to `builder:22` seems reasonable.

runesoerensen: I took a stab at updating the readme to better reflect the language/buildpack support per builder (at the expense of the Default Run Image and Lifecycle Version values, which I suspect few will/have ever needed a link for here).

edmorley: The reason I included the lifecycle values, is that it affects supports Buildpack API versions. ie: When we update lifecycle it can break older buildpacks. The later part of the README then references the lifecycle version:
https://github.com/heroku/cnb-builder-images/blob/f244fcf83610453ff772401ac41818331ee611a4/README.md?plain=1#L27-L28

There's quite possibly a better way to do this, eg:
- now we're using the same lifecycle version for all buildpacks, perhaps that part can be moved out of the table?
- perhaps we should list "Supported Buildpack API versions" instead of lifecycle versions? Or in addition to? (Given that there is also the list of supported Platform API versions, which impacts what Pack CLI or Kodon or kpack etc versions will work with the builder)

If we do end up removing the lifecycle version from the table, then at a minimum we'll need to:
- adjust https://github.com/heroku/cnb-builder-images/blob/f244fcf83610453ff772401ac41818331ee611a4/README.md?plain=1#L27-L28
- adjust this workflow: https://github.com/heroku/cnb-builder-images/blob/f244fcf83610453ff772401ac41818331ee611a4/.github/workflows/update-lifecycle.yml#L45-L46

edmorley: Also the reason the run image was in the table wasn't just because of quick links to tags, but to help with users understanding the relationship between the images, added in #440 as an alternative/compromise to #434 / #437.

(Again, open to ideas and there are quite possibly better ways to do this; just explaining the context :-) )

runesoerensen: Very much appreciate the context. And I certainly see the value in communicating the lifecycle version, as well as linking to the run image. I didn't mean for the last readme update to be a final version :) But yes since we're (currently) using the same lifecycle that does seem like an item we could potentially move out of the table (if we want/need to make space for of supported buildpacks)

edmorley: > * now we're using the same lifecycle version for all buildpacks,

More context: Having all documented images be on the same lifecycle version occurred via:
- https://github.com/heroku/cnb-builder-images/pull/514
- the functions image (which are on an older lifecycle) now being a separate thing, and (intentionally) not documented in the README

I imagine we'll end up with a mix of lifecycle versions again when we EOL `heroku/builder:20` - at which point the configs and workflows for that builder will be deleted from the repo (so the images won't be updated any more), but we'll presumably still list it in the README table for a while as "end of life". (I suppose we could just delete it from the table at that point, or create a second "EOL images" table?)

runesoerensen: Would it maybe be better to revert the README changes in this PR, and instead simply adjust the phrasing of the sentence following the table in a way that doesn't call out .NET specifically.

Based on feedback so far, it sounds like it'd probably make sense to add `heroku/dotnet` to both `heroku/builder:22` and `heroku/builder:24`. Those are the current and recommended builders - how about we just replace the mention of supported languages with a link to https://github.com/heroku/buildpacks?tab=readme-ov-file#supported-languages?

edmorley: If we're not adding dotnet to `heroku/builder:20` (which I can see the reasoning behind, and am fine with), then I'm fine with adding the README .NET "Additionally, XYZ includes dotnet..." wording back (that was in an earlier version of the PR), and then revisiting the README in general at a later date.

> how about we just replace the mention of supported languages with a link to https://github.com/heroku/buildpacks?tab=readme-ov-file#supported-languages?

I suppose some downsides of that are (a) it's an extra click/easier to get out of sync, (b) in the future there might be official buildpacks that are "supported" but not in the builder.

runesoerensen: Alright, I'm ok with adding the original wording back (updated to reflect `heroku/builder:22` support), and then revisiting at a later date. https://github.com/heroku/cnb-builder-images/pull/571/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R26

edmorley: Yeah now we have https://github.com/heroku/buildpacks perhaps that's the best place for overview/concept stuff like how builder, run image, buildpacks etc intersect.

edmorley (approved):
🎉

PR #591, opened 2024-10-11
Add `heroku/deb-packages` to 22/24 builders

colincasey:
Registers the [heroku/deb-packages buildpack](https://github.com/heroku/buildpacks-deb-packages) in the following builders, prepending it to each buildpack group as an optional buildpack that runs at the start for:
- `heroku/builder:24`
- `heroku/builder:22`

> [!note]
> This will not be added to `heroku/builder:20` as that builder is deprecated.

GUS-W-17066441.

### DIFF
--- a/builder-22/builder.toml
+++ b/builder-22/builder.toml
@@ -18,6 +18,10 @@ mirrors = ["public.ecr.aws/heroku/heroku:22-cnb"]
 version = "0.20.3"
 
 [[buildpacks]]
+  id = "heroku/deb-packages"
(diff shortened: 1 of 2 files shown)


edmorley (approved):
Thank you for adding this to the builder - I think having it included is a great win for UX.

I notice the buildpack wasn't added to `heroku/builder:20` - I think this makes sense, but is perhaps worth calling out in the PR description to avoid user confusion.

Medium/longer term I'd love to have upstream support for system buildpacks be implemented (the RFC for it was accepted some time ago, but it hasn't been implemented yet: https://github.com/buildpacks/rfcs/issues/209) - since both `heroku/procfile` and now `heroku/deb-packages` could make use of it :-)

joshwlewis (approved):

PR #552, opened 2024-07-17
Add deprecation messaging to builder:20

joshwlewis:
This adds a deprecation message to all builds using `heroku/builder:20`. 

Since this builder is based on the now deprecated `heroku/heroku:20` ([changelog entry](https://devcenter.heroku.com/changelog-items/2895)), security updates for `heroku/builder:20` will stop April 30, 2025. This messaging should encourage folks to update to a newer builder so that they may keep getting security updates and avoid interruption.

Example pack output:
Pack CLI output

[heroku-20 EOL plan]( https://salesforce.quip.com/ss13A0jDWb7a)
[GUS](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001gotTdYAI/view)

PR #712, opened 2025-04-29
Upgrade the `heroku/builder:20` deprecation warning to an error

edmorley:
Heroku-20, Heroku's Ubuntu 20.04 based stack, was previously deprecated in June 2024:
https://devcenter.heroku.com/changelog-items/2895

The `heroku/builder:20` image has been outputting deprecation warnings since July 2024:
https://github.com/heroku/cnb-builder-images/pull/552

Now, in accordance with our stack support policy and the documented EOL timelines, the stack and thus the builder image have reached end-of-life.

As such, the EOL warnings have been upgraded to errors (which can be bypassed by setting an env var), similar to what we did for the legacy builder image sunset in #478.

Users who want to continue to use this builder, will need to set the env var:
`ALLOW_EOL_HEROKU_BUILDER_20=1`

The EOL warning buildpack has been moved first in the group detection order, so that builds fail fast - preventing wasting time and also making the error clearly since it's not surrounded by the rest of the buildpack logs.

GUS-W-14706656.

### DIFF
--- a/.github/workflows/build-test-publish.yml
+++ b/.github/workflows/build-test-publish.yml
@@ -99,7 +99,7 @@ jobs:
       - name: Load Docker images into the Docker daemon
         run: zstd -dc --long=31 images.tar.zst | docker load
       - name: Build getting started guide image
-        run: pack build getting-started --force-color --builder ${{ matrix.builder }} --trust-builder --pull-policy never
+        run: pack build getting-started --force-color --builder ${{ matrix.builder }} --trust-builder --pull-policy never --env ALLOW_EOL_HEROKU_BUILDER_20=1
       - name: Start getting started guide image
         run: docker run --name getting-started --detach -p 8080:8080 --env PORT=8080 getting-started
       - name: Test getting started web server response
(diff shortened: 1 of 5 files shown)


edmorley:
An example failing build when the env var is not set:
https://github.com/heroku/cnb-builder-images/actions/runs/14726043050/job/41328874535?pr=712#step:6:1

And a successful build when `ALLOW_EOL_HEROKU_BUILDER_20=1` is set (that still emits the EOL warning):
https://github.com/heroku/cnb-builder-images/actions/runs/14726084940/job/41329029738#step:6:1

This PR won't be merged until May 1st.

schneems (approved):

.github/workflows/build-test-publish.yml
```diff
@@ -99,7 +99,7 @@ jobs:
       - name: Load Docker images into the Docker daemon
         run: zstd -dc --long=31 images.tar.zst | docker load
       - name: Build getting started guide image
-        run: pack build getting-started --force-color --builder ${{ matrix.builder }} --trust-builder --pull-policy never
+        run: pack build getting-started --force-color --builder ${{ matrix.builder }} --trust-builder --pull-policy never --env ALLOW_EOL_HEROKU_BUILDER_20=1
```
schneems, line 102: > This PR won't be merged until May 1st.

This comment will prevent someone from accidentally merging sooner

edmorley:
https://devcenter.heroku.com/changelog-items/3230

PR #898, opened 2026-01-29
Add Front-end Web CNB into buildpack detection sequence

mars:
In preparation for launch of the Front-end Web CNB, let's start discussing how to add it into the buildpack detection sequence. This is based on the [approved, internal RFC](https://github.com/heroku/rfcs/blob/main/approved/000055-Front-end_Web_Buildpack.md#detect).

### DIFF
--- a/builder-24/builder.toml
+++ b/builder-24/builder.toml
@@ -35,7 +35,7 @@ mirrors = ["public.ecr.aws/heroku/heroku:24"]
 
 [[buildpacks]]
   id = "heroku/nodejs"
-  uri = "docker://docker.io/heroku/buildpack-nodejs@sha256:57e70688c97bca0e7fc63f4afcfd0fb7e67c5d267a5db1860e75331835ae6b75"
+  uri = "docker://docker.io/heroku/buildpack-nodejs@sha256:41d5c88d8605a84f9ae464c869d40705b6d5cb1b5ea83702ce05a8def41cc5c2"
 
 [[buildpacks]]
   id = "heroku/php"
@@ -97,7 +97,7 @@ mirrors = ["public.ecr.aws/heroku/heroku:24"]
     optional = true
   [[order.group]]
     id = "heroku/nodejs"
-    version = "5.4.2"
+    version = "5.4.3"
     optional = true
   [[order.group]]
     id = "heroku/jvm"
@@ -244,7 +244,7 @@ mirrors = ["public.ecr.aws/heroku/heroku:24"]
     optional = true
   [[order.group]]
     id = "heroku/nodejs"
-    version = "5.4.2"
+    version = "5.4.3"
   [[order.group]]
     id = "heroku/procfile"
     version = "4.2.1"
--- a/builder-24/builder.toml
+++ b/builder-24/builder.toml
@@ -1,4 +1,4 @@
-description = "Ubuntu 24.04 AMD64+ARM64 base image with buildpacks for .NET, Go, Java, Node.js, PHP, Python, Ruby & Scala."
+description = "Ubuntu 24.04 AMD64+ARM64 base image with buildpacks for .NET, Go, Java, Node.js, PHP, Python, Ruby, Scala, & front-end websites."
 
 [stack]
 id = "heroku-24"
--- a/builder-24/builder.toml
+++ b/builder-24/builder.toml
@@ -47,7 +47,7 @@ mirrors = ["public.ecr.aws/heroku/heroku:24"]
 
 [[buildpacks]]
   id = "heroku/python"
-  uri = "docker://docker.io/heroku/buildpack-python@sha256:4f15bd861d5f7bc710b43c63a182f66c319771a76fea471a18bd77808ba5a21a"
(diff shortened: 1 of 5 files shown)


builder-24/builder.toml
```diff
@@ -57,6 +57,14 @@ mirrors = ["public.ecr.aws/heroku/heroku:24"]
   id = "heroku/scala"
   uri = "docker://docker.io/heroku/buildpack-scala@sha256:0d25c94222de9f64de1a43b5c9ee7645c82dec7b4d8620f6e45294d345a3b3cd"
 
+[[buildpacks]]
+  id = "heroku/website-nodejs"
```
edmorley: One potential issue that occurred to me, is that with this change, we'll now have `heroku/nodejs` pulled in via two separate paths that may not use the same version, and I seem to remember Pack CLI errors if it can't disambiguate which CNB version to use.

ie Imagine:
- This PR merges
- Node.js CNB releases a new version, and the automation updates the direct references in this repo to use the `heroku/nodejs` new version
- However, the builder image now contains two versions of `heroku/nodejs` - those that are direct references in the manifest in this repo (from the Node.js support and the Ruby optional Node.js feature), plus the transitive `heroku/nodejs` version pulled in via `heroku/website-nodejs` (which at least initially, will still be using the old `heroku/nodejs` version until its updated over there too and a new release of the website CNBs cut)
- When someone does a `pack build --buildpack heroku/nodejs` (or has the `heroku/nodejs` CNB configured via their `project.toml`), ordinarily Pack CLI uses the CNB version from the builder image (if it exists there). This has several advantages like ensuring the builder image is deterministic in terms of buildpack versions used, but also means the common case (of people not specifying an exact `heroku/nodejs` version) doesn't have to make additional requests to Docker Hub (avoiding rate limit issues) since it uses the bundles CNB assets rather than pulling additional CNB images.
- However, now that there are multiple `heroku/nodejs` CNBs of different versions in the builder, what does Pack CLI use? I seem to remember it errors, but we should check. Plus does how does Kodon handle it? (buildpack order handling is one of those few things that has to exist outside of `lifecycle`, so has to be implemented in the "platform", so has separate implementations in Pack CLI vs Kodon vs ... etc)

If Pack (and/or Kodon) do error, then we'll have to either:
(a) always try and keep `heroku/website-nodejs`'s transitive dependency on `heroku/nodejs` in sync with the direct dependency versions in this repo (which is going to break all the automation unless we change the approach completely, and at the very least require manual PRs for Colin short term)
(b) work with upstream to come up with a better way for this to work. For example ideally `heroku/website-nodejs` could specify a wildcard or semver range or something?
(c) think about moving the composite buildpack definition into this repo, like was done for Ruby's optional Node.js feature (to avoid this issue)

edmorley (commented):
Thank you for opening the PR!

colincasey (commented):
I think my main concern is if the detection logic for certain front-end buildpacks might be too broad. For example, `vite` is less of a framework and more of a tool to support frameworks. Could triggering that detection capture apps that may not want the front-end buildpack behavior? Maybe the answer is it'll never be perfect and documentation is needed for those edge cases?

builder-24/builder.toml
mars: This direction follows @colincasey's comment later in the review:

> I also wonder if instead of including the [heroku/website-nodejs composite buildpack](https://github.com/heroku/buildpacks-frontend-web/blob/main/meta-buildpacks/website-nodejs/buildpack.toml), the builder should explicitly list out the frameworks.

mars:
> For example, `vite` is less of a framework and more of a tool to support frameworks.

This is a fair point… Vite provides a [default build output directory](https://vite.dev/config/build-options#build-outdir), so really all `website-vite` does is configure that output as the web server root and expect a an `index.html`.

If this seems to broad, then it's possible we could do deeper Vite config detection by executing Node.js during detect to evaluate the [`vite.config.js`](https://vite.dev/config/) source. I imagine it is not possible to exec Node during detect, though.

Or maybe, during the build phase, when Node.js is available, evaluate the Vite config to dynamically adapt web server configuration.

colincasey (commented):
I'm 👍 for the changes here (and https://github.com/heroku/buildpacks-frontend-web/pull/68). 

One last thing though - it may seem redundant but adding a smoke test for each supported front-end framework would be beneficial here. There may not be a specific "Getting Started" app from Heroku for these but each framework typically has their own "Getting Started" template you could use. 

The tests here complement the integration tests in our individual buildpacks by validating that:
* all buildpacks combinations work together
* detection groups are operating as expected

This has helped to identify issues like [this one](https://github.com/heroku/cnb-builder-images/pull/797) from when I refactored the Node.js CNB to collapse the component buildpacks into a single buildpack + and replaced the composite buildpack with that single buildpack. The tests here showed what my repo's integration tests couldn't - that there were incompatibilities with the Ruby CNB that needed to be addressed.

edmorley:
Closing this PR for now since it has several conflicts, doesn't have the new Heroku-26 stack, and would need updates to the versions of all the other CNBs such as nodejs and procfile etc. If/when we're ready to add this CNB to the builder we can either reopen and rebase, or create a new PR :-)

PR #979, opened 2026-05-21
Auto-approve buildpack update PRs

runesoerensen:
Approves PRs opened by `heroku-linguist[bot]` from `update/heroku/buildpacks-*` branches using the workflow's `GITHUB_TOKEN`, so the auto-merge already enabled by the buildpack release workflow can fire without a human approval step. Branch protection on `main` continues to enforce the security boundary.

Triple-gated: same-repo (rejects fork PRs), bot author, branch-name prefix. `synchronize` is omitted so approval only fires on initial PR open.

**Requires** "Allow GitHub Actions to create and approve pull requests" to be enabled in [repo settings](https://github.com/heroku/cnb-builder-images/settings/actions) before this is effective.

GUS-W-22597481.

### DIFF
--- a/.github/workflows/auto-approve-buildpack-updates.yml
+++ b/.github/workflows/auto-approve-buildpack-updates.yml
@@ -21,6 +21,8 @@ jobs:
     # rejects API calls from GitHub-hosted runners.
     runs-on: pub-hk-ubuntu-24.04-ip
     steps:
+      # Approves as `github-actions[bot]` rather than the Linguist user,
+      # since a user can't approve their own PR.
       - name: Approve PR
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/auto-approve-buildpack-updates.yml
+++ b/.github/workflows/auto-approve-buildpack-updates.yml
@@ -1,0 +1,28 @@
+name: Auto-approve buildpack update PRs
+
+# Uses `pull_request` (not `pull_request_target`), so fork-source PRs run with
+# secrets stripped and a read-only `GITHUB_TOKEN`. `synchronize` is omitted,
+# so approval only fires on the initial PR open.
+on:
+  pull_request:
+    types:
+      - opened
+
+permissions:
+  pull-requests: write
+
+jobs:
+  auto-approve:
+    if: >-
+      github.event.pull_request.head.repo.full_name == github.repository &&
+      github.event.pull_request.user.login == 'heroku-linguist[bot]' &&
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-approve-buildpack-updates.yml
```diff
@@ -0,0 +1,28 @@
...
+      github.event.pull_request.head.repo.full_name == github.repository &&
+      github.event.pull_request.user.login == 'heroku-linguist[bot]' &&
+      startsWith(github.event.pull_request.head.ref, 'update/heroku/buildpacks-')
+    # IP-allowlisted runner used because the heroku org's IP allow list
+    # rejects API calls from GitHub-hosted runners.
+    runs-on: pub-hk-ubuntu-24.04-ip
+    steps:
+      - name: Approve PR
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
edmorley, line 33: (optional) Do we need a comment saying we're using this token and not the Linguist user since a user can't approve their own PRs?

.github/workflows/auto-approve-buildpack-updates.yml
```diff
@@ -0,0 +1,28 @@
+name: Auto-approve buildpack update PRs
+
+# Uses `pull_request` (not `pull_request_target`), so fork-source PRs run with
+# secrets stripped and a read-only `GITHUB_TOKEN`. `synchronize` is omitted,
+# so approval only fires on the initial PR open.
+on:
```
edmorley, line 6: (optional) Do we want to use the paths filter so this doesn't run at all for PRs that only touch eg GitHub Actions workflow (eg Dependabot PRs)? Whilst the `if` will stop it actually fully executing, it will still partially trigger and cause a job to be executed (vs the path filters which aiui kick in before the workflow is even queued)

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpushpull_requestpull_request_targetpathspaths-ignore

edmorley (approved):
Amazing, thank you!

.github/workflows/auto-approve-buildpack-updates.yml
runesoerensen: Done 2aaa0a00910ea12192e16f459c73944b126746ff

.github/workflows/auto-approve-buildpack-updates.yml
runesoerensen: Good idea -- I've added a filter on `builder.toml` paths here ba7e2f051d0351212921ea7ae05d716f79ef8610

PR #989, opened 2026-06-05
Post Honeycomb release marker on builder image publish

colincasey:
Posts a release marker to the Honeycomb `heroku-builds-production` dataset when the builder images are published.

## What

A new `post-honeycomb-marker` job (`needs: publish-manifest`) that posts a [marker](https://docs.honeycomb.io/configure/environments/manage-markers/) to the Honeycomb `heroku-builds-production` dataset on a successful builder release, via the [create-a-marker API](https://docs.honeycomb.io/api/markers/create-a-marker/).

## Behavior

- **Every release:** the job is gated on `success() && github.event_name == 'push' && github.ref == 'refs/heads/main'`. Every push to `main` rebuilds and republishes the builder (picking up new base images, buildpack updates, detection-order changes, etc.), so each one is a genuine release worth tracking. Excludes the nightly `schedule` rebuild and `workflow_dispatch`, both of which republish unchanged images.
- **Idempotent:** dedups on the commit URL, so a workflow re-run finds the existing marker and skips rather than stacking a duplicate.
- **Anchored to the release:** `start_time` is set from the release commit timestamp, so a delayed or re-triggered run still anchors the marker to the release rather than to job-execution time.
- **Fails loud:** this is the final job and runs after the release has already published, so a Honeycomb API error fails the job red (surfacing a retry/bug) without affecting the release.

## Follow-up

Requires the `HONEYCOMB_API_KEY` repo secret (env API key with the create-markers scope) to be added before markers will post.

W-22860436

### DIFF
--- a/.github/workflows/build-test-publish.yml
+++ b/.github/workflows/build-test-publish.yml
@@ -217,3 +217,64 @@ jobs:
           set -x
           docker manifest create "${{ matrix.tag_uri }}" ${{ matrix.manifest_uris }}
           docker manifest push "${{ matrix.tag_uri }}"
+
+  post-honeycomb-marker:
+    name: Post Honeycomb release marker
+    needs: [publish-image, publish-manifest]
+    # Gate to genuine releases only: a push to main, never the nightly `schedule`
+    # rebuild or a `workflow_dispatch`, both of which republish unchanged images.
+    # The buildpack-update check below narrows it further.
+    if: success() && github.event_name == 'push' && github.ref == 'refs/heads/main'
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+        # Only the tip commit's author + subject are needed to confirm + label the release.
+        with:
+          fetch-depth: 1
+
+      - name: Post Honeycomb release marker
+        # No-op if the secret isn't configured (e.g. before it's added to the repo),
+        # rather than failing an otherwise-successful release. Secrets can't be used
+        # in an `if:`, so the key is mirrored to env and tested there — GitHub masks
+        # registered secrets in logs regardless, so the mirror is safe.
+        if: env.HONEYCOMB_API_KEY
+        env:
+          HONEYCOMB_API_KEY: ${{ secrets.HONEYCOMB_API_KEY }}
+          HONEYCOMB_DATASET: builds
+          COMMIT_URL: ${{ github.server_url }}/${{ github.repository }}/commit/${{ github.sha }}
+        run: |
+          msg=$(git log -1 --format=%s)
+          author=$(git log -1 --format=%an)
+
+          # Builder releases reach main as a Linguist buildpack-update merge, authored by
+          # heroku-linguist[bot] and titled "Update heroku/buildpacks-<lang> to vX.Y.Z" by
+          # the release automation. Require both: the bot author excludes hand-pushed
+          # commits, and the subject excludes the bot's own lifecycle bumps ("Update
+          # lifecycle from ..."). Anything else reaches main too but isn't a release —
+          # skip it without failing.
+          if [[ "$author" != "heroku-linguist[bot]" || "$msg" != Update\ heroku/buildpacks-* ]]; then
+            echo "Not a buildpack-update release (author=${author}, subject=${msg}) — skipping marker."
+            exit 0
+          fi
+
+          api="https://api.honeycomb.io/1/markers/${HONEYCOMB_DATASET}"
+          body=$(jq -n --arg msg "$msg" --arg url "$COMMIT_URL" '{message: $msg, type: "builder-release", url: $url}')
+          auth=(-H "X-Honeycomb-Team: ${HONEYCOMB_API_KEY}" -H "Content-Type: application/json")
+
+          # One marker per builder release; dedup on the commit URL (stable identity for
+          # this release) so a re-run of the workflow updates the existing marker rather
+          # than stacking a duplicate. Every path below warns on failure and never exits
+          # non-zero — a Honeycomb outage must not fail an already-successful release.
+          existing=$(curl -sS --retry 3 --retry-all-errors --max-time 30 "${auth[@]}" "${api}" \
+            | jq -r --arg url "$COMMIT_URL" 'map(select(.url == $url)) | first | .id // empty' 2>/dev/null) || existing=""
+
+          if [[ -n "$existing" ]]; then
+            curl -sS --retry 3 --retry-all-errors --max-time 30 -X PUT "${api}/${existing}" "${auth[@]}" -d "${body}" \
+              >/dev/null && echo "Updated existing release marker (${existing})" \
+              || echo "::warning::Failed to update Honeycomb release marker (release succeeded regardless)"
+          else
+            curl -sS --retry 3 --retry-all-errors --max-time 30 -X POST "${api}" "${auth[@]}" -d "${body}" \
+              >/dev/null && echo "Created release marker for ${msg}" \
+              || echo "::warning::Failed to create Honeycomb release marker (release succeeded regardless)"
+          fi
--- a/.github/workflows/build-test-publish.yml
+++ b/.github/workflows/build-test-publish.yml
@@ -259,22 +259,24 @@ jobs:
           fi
 
           api="https://api.honeycomb.io/1/markers/${HONEYCOMB_DATASET}"
-          body=$(jq -n --arg msg "$msg" --arg url "$COMMIT_URL" '{message: $msg, type: "builder-release", url: $url}')
-          auth=(-H "X-Honeycomb-Team: ${HONEYCOMB_API_KEY}" -H "Content-Type: application/json")
+          curl=(curl -sS --fail --connect-timeout 10 --retry 3 --retry-all-errors --max-time 30
+            -H "X-Honeycomb-Team: ${HONEYCOMB_API_KEY}" -H "Content-Type: application/json")
 
           # One marker per builder release; dedup on the commit URL (stable identity for
-          # this release) so a re-run of the workflow updates the existing marker rather
-          # than stacking a duplicate. Every path below warns on failure and never exits
-          # non-zero — a Honeycomb outage must not fail an already-successful release.
-          existing=$(curl -sS --retry 3 --retry-all-errors --max-time 30 "${auth[@]}" "${api}" \
-            | jq -r --arg url "$COMMIT_URL" 'map(select(.url == $url)) | first | .id // empty' 2>/dev/null) || existing=""
+          # this release). If a marker already exists (e.g. a re-run of the workflow), skip
+          # — the marker is unchanged, so there's nothing to update. Every call warns on
+          # failure and never exits non-zero: a Honeycomb outage must not fail an
+          # already-successful release.
+          existing=$("${curl[@]}" "${api}" \
+            | jq -r --arg url "$COMMIT_URL" 'map(select(.url == $url)) | first | .id // empty') \
+            || { echo "::warning::Failed to query Honeycomb markers (release succeeded regardless)"; exit 0; }
 
           if [[ -n "$existing" ]]; then
-            curl -sS --retry 3 --retry-all-errors --max-time 30 -X PUT "${api}/${existing}" "${auth[@]}" -d "${body}" \
-              >/dev/null && echo "Updated existing release marker (${existing})" \
-              || echo "::warning::Failed to update Honeycomb release marker (release succeeded regardless)"
-          else
-            curl -sS --retry 3 --retry-all-errors --max-time 30 -X POST "${api}" "${auth[@]}" -d "${body}" \
-              >/dev/null && echo "Created release marker for ${msg}" \
-              || echo "::warning::Failed to create Honeycomb release marker (release succeeded regardless)"
+            echo "Release marker already exists (${existing}) — skipping."
+            exit 0
           fi
+
+          body=$(jq -n --arg msg "$msg" --arg url "$COMMIT_URL" '{message: $msg, type: "builder-release", url: $url}')
+          "${curl[@]}" -X POST "${api}" -d "${body}" >/dev/null \
+            && echo "Created release marker for ${msg}" \
+            || echo "::warning::Failed to create Honeycomb release marker (release succeeded regardless)"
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test-publish.yml
```diff
@@ -217,3 +217,64 @@ jobs:
...
+          api="https://api.honeycomb.io/1/markers/${HONEYCOMB_DATASET}"
+          body=$(jq -n --arg msg "$msg" --arg url "$COMMIT_URL" '{message: $msg, type: "builder-release", url: $url}')
+          auth=(-H "X-Honeycomb-Team: ${HONEYCOMB_API_KEY}" -H "Content-Type: application/json")
+
+          # One marker per builder release; dedup on the commit URL (stable identity for
+          # this release) so a re-run of the workflow updates the existing marker rather
+          # than stacking a duplicate. Every path below warns on failure and never exits
+          # non-zero — a Honeycomb outage must not fail an already-successful release.
+          existing=$(curl -sS --retry 3 --retry-all-errors --max-time 30 "${auth[@]}" "${api}" \
+            | jq -r --arg url "$COMMIT_URL" 'map(select(.url == $url)) | first | .id // empty' 2>/dev/null) || existing=""
```
edmorley: I don't think we should pipe errors to `/dev/null` - let them be printed as is (only the stdout gets captured by the `$()` so the stderr is fine to still print, since it will be shown directly in the logs)

.github/workflows/build-test-publish.yml
```diff
@@ -217,3 +217,64 @@ jobs:
...
+
+          if [[ -n "$existing" ]]; then
+            curl -sS --retry 3 --retry-all-errors --max-time 30 -X PUT "${api}/${existing}" "${auth[@]}" -d "${body}" \
+              >/dev/null && echo "Updated existing release marker (${existing})" \
+              || echo "::warning::Failed to update Honeycomb release marker (release succeeded regardless)"
+          else
+            curl -sS --retry 3 --retry-all-errors --max-time 30 -X POST "${api}" "${auth[@]}" -d "${body}" \
```
edmorley: Could you add the missing standard curl best practices like `--fail` and `--connect-timeout`? (Without `--fail`, we won't get the annotation when the API returns a non-200 value etc)

.github/workflows/build-test-publish.yml
```diff
@@ -217,3 +217,64 @@ jobs:
...
+          # this release) so a re-run of the workflow updates the existing marker rather
+          # than stacking a duplicate. Every path below warns on failure and never exits
+          # non-zero — a Honeycomb outage must not fail an already-successful release.
+          existing=$(curl -sS --retry 3 --retry-all-errors --max-time 30 "${auth[@]}" "${api}" \
+            | jq -r --arg url "$COMMIT_URL" 'map(select(.url == $url)) | first | .id // empty' 2>/dev/null) || existing=""
+
+          if [[ -n "$existing" ]]; then
+            curl -sS --retry 3 --retry-all-errors --max-time 30 -X PUT "${api}/${existing}" "${auth[@]}" -d "${body}" \
+              >/dev/null && echo "Updated existing release marker (${existing})" \
+              || echo "::warning::Failed to update Honeycomb release marker (release succeeded regardless)"
```
edmorley: Do we want to be updating the entry if an existing one exists? It feels like we can/should just skip in that case?

In which case this also means the implementation here can be written with much less duplication.

.github/workflows/build-test-publish.yml
```diff
@@ -217,3 +217,64 @@ jobs:
...
+          # Builder releases reach main as a Linguist buildpack-update merge, authored by
+          # heroku-linguist[bot] and titled "Update heroku/buildpacks-<lang> to vX.Y.Z" by
+          # the release automation. Require both: the bot author excludes hand-pushed
+          # commits, and the subject excludes the bot's own lifecycle bumps ("Update
+          # lifecycle from ..."). Anything else reaches main too but isn't a release —
+          # skip it without failing.
+          if [[ "$author" != "heroku-linguist[bot]" || "$msg" != Update\ heroku/buildpacks-* ]]; then
+            echo "Not a buildpack-update release (author=${author}, subject=${msg}) — skipping marker."
+            exit 0
+          fi
```
edmorley: If a human makes a change (such as adding a new buildpack, or changing the detection order) wouldn't it still be useful to have the release marker? 

ie: Can this check just be removed, and we emit markers for everything? Since any time a push happens on main we rebuild the builder, and it picks up any changes, such as new base images - even if the commit here were to just eg the README.

edmorley (changes requested):

.github/workflows/build-test-publish.yml
colincasey: @edmorley i suppose i was just worried about creating noise with unrelated release changes. maybe this should be gated on if there was a change to `builder-XX/*` files? that way README, workflow changes, etc. won't create a marker. what do you think?

.github/workflows/build-test-publish.yml
colincasey: Fixed in f78b43f

.github/workflows/build-test-publish.yml
colincasey: Fixed in f78b43f

.github/workflows/build-test-publish.yml
colincasey: Fixed in f78b43f

.github/workflows/build-test-publish.yml
edmorley: @colincasey That was my first thought, however, bear in mind:

> Since any time a push happens on main we rebuild the builder, and it picks up any changes, such as new base images - even if the commit here were to just eg the README.

ie: Any commit causes a new image to go out. And if there is a new image, then we should track it.

.github/workflows/build-test-publish.yml
```diff
@@ -217,3 +217,66 @@ jobs:
...
+        # Only the tip commit's author + subject are needed to confirm + label the release.
+        with:
+          fetch-depth: 1
+
+      - name: Post Honeycomb release marker
+        # No-op if the secret isn't configured (e.g. before it's added to the repo),
+        # rather than failing an otherwise-successful release. Secrets can't be used
+        # in an `if:`, so the key is mirrored to env and tested there — GitHub masks
+        # registered secrets in logs regardless, so the mirror is safe.
+        if: env.HONEYCOMB_API_KEY
```
edmorley: This seems unnecessary - we could remove this and just make sure the env var is added before this PR is merged? (Which would also avoid having to have a 4 line comment explaining it)

.github/workflows/build-test-publish.yml
colincasey: dropped the author/subject filter in 9e86d79

.github/workflows/build-test-publish.yml
colincasey: removed the guard and its comment in 9e86d79

.github/workflows/build-test-publish.yml
```diff
@@ -217,3 +217,49 @@ jobs:
           set -x
           docker manifest create "${{ matrix.tag_uri }}" ${{ matrix.manifest_uris }}
           docker manifest push "${{ matrix.tag_uri }}"
+
+  post-honeycomb-marker:
+    name: Post Honeycomb release marker
+    needs: [publish-image, publish-manifest]
+    # Every push to main rebuilds and republishes the builder, so each one is a genuine
+    # release worth tracking. Excludes the nightly `schedule` rebuild and `workflow_dispatch`,
+    # both of which republish unchanged images.
```
edmorley, line 226: Could you update the PR description/commit message to match the new behaviour? (That markers are posted for every push, not just buildpack update PRs)

Also, whilst reviewing the PR I had to manually look up the Honeycomb Markers API docs - for posterity could you link to https://docs.honeycomb.io/configure/environments/manage-markers and https://docs.honeycomb.io/api/markers/create-a-marker in the PR description too?

.github/workflows/build-test-publish.yml
```diff
@@ -217,3 +217,49 @@ jobs:
...
+            -H "X-Honeycomb-Team: ${HONEYCOMB_API_KEY}" -H "Content-Type: application/json")
+
+          # One marker per builder release; dedup on the commit URL (stable identity for
+          # this release). If a marker already exists (e.g. a re-run of the workflow), skip
+          # — the marker is unchanged, so there's nothing to update. Every call warns on
+          # failure and never exits non-zero: a Honeycomb outage must not fail an
+          # already-successful release.
+          existing=$("${curl[@]}" "${api}" \
+            | jq -r --arg url "$COMMIT_URL" 'map(select(.url == $url)) | first | .id // empty') \
+            || { echo "::warning::Failed to query Honeycomb markers (release succeeded regardless)"; exit 0; }
```
edmorley: So thinking about this some more, I think it might be better if we removed the `|| { echo "::warning::` usages - I think we'd rather see a red job here, so we know to retry it (or that there's a bug) rather than just have the warning.

Since this is the last step in the workflow, having a red job won't prevent the actual release, so there doesn't seem to be any downside of doing that?

.github/workflows/build-test-publish.yml
```diff
@@ -217,3 +217,49 @@ jobs:
...
+
+          if [[ -n "$existing" ]]; then
+            echo "Release marker already exists (${existing}) — skipping."
+            exit 0
+          fi
+
+          body=$(jq -n --arg msg "$msg" --arg url "$COMMIT_URL" '{message: $msg, type: "builder-release", url: $url}')
+          "${curl[@]}" -X POST "${api}" -d "${body}" >/dev/null \
```
edmorley: In general I think we should avoid piping to `/dev/null` otherwise it can make debugging harder?

.github/workflows/build-test-publish.yml
```diff
@@ -217,3 +217,49 @@ jobs:
           set -x
           docker manifest create "${{ matrix.tag_uri }}" ${{ matrix.manifest_uris }}
           docker manifest push "${{ matrix.tag_uri }}"
+
+  post-honeycomb-marker:
+    name: Post Honeycomb release marker
+    needs: [publish-image, publish-manifest]
```
edmorley: `publish-manifest` already depends on `publish-image` - the job sequence here is linear

```suggestion
    needs: publish-manifest
```

.github/workflows/build-test-publish.yml
```diff
@@ -217,3 +217,49 @@ jobs:
...
+          existing=$("${curl[@]}" "${api}" \
+            | jq -r --arg url "$COMMIT_URL" 'map(select(.url == $url)) | first | .id // empty') \
+            || { echo "::warning::Failed to query Honeycomb markers (release succeeded regardless)"; exit 0; }
+
+          if [[ -n "$existing" ]]; then
+            echo "Release marker already exists (${existing}) — skipping."
+            exit 0
+          fi
+
+          body=$(jq -n --arg msg "$msg" --arg url "$COMMIT_URL" '{message: $msg, type: "builder-release", url: $url}')
```
edmorley: In https://github.com/heroku/languages-github-actions/pull/369 the `type` is set to `deploy`, vs `builder-release` here. I think we should align both to be consistent? (Currently one is more specific than the other)

What do the `types` actually do?

edmorley: Should we be setting `start_time` explicitly to the time of the publish? (Since this step may run delayed after the fact, particularly if it had to be retriggered etc)

https://docs.honeycomb.io/api/markers/create-a-marker#body-start-time

edmorley (dismissed):
Thank you for the updates!

I've left some more comments, but I'm approving this so you can merge after making any necessary changes, given I'm away from tomorrow until Wed :-)

In addition to the comments here - could you also cross-check this PR with https://github.com/heroku/languages-github-actions/pull/369 in general and make sure the two are consistent (eg to ensure a fix/change that applies to both has been made to both).

.github/workflows/build-test-publish.yml
```diff
@@ -217,3 +217,49 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v6
+        # Only the tip commit's subject is needed to label the release.
+        with:
+          fetch-depth: 1
+
+      - name: Post Honeycomb release marker
+        env:
+          HONEYCOMB_API_KEY: ${{ secrets.HONEYCOMB_API_KEY }}
+          HONEYCOMB_DATASET: builds
```
edmorley: Is this the right dataset? I thought CNBs and classic used different datasets?

.github/workflows/build-test-publish.yml
colincasey: Done — uses `needs: publish-manifest` now.

.github/workflows/build-test-publish.yml
colincasey: Good catch — this was wrong, it's `heroku-builds-production` for CNBs.

.github/workflows/build-test-publish.yml
colincasey: Aligned to `deploy` to match #369. The `type` groups markers and lets you assign a per-type color in the dataset settings — no behavioral effect beyond display.

colincasey: Added — set from the commit timestamp (`git log -1 --format=%ct`) rather than "now", so a delayed/re-triggered run still anchors the marker to the release.

.github/workflows/build-test-publish.yml
colincasey: Done — dropped the `::warning::` fallbacks and the `/dev/null` redirects. With `--fail` and bash's `pipefail`, a Honeycomb error now fails this (last) job red without affecting the already-published release.

.github/workflows/build-test-publish.yml
colincasey: Done — dropped the `/dev/null` redirects.

.github/workflows/build-test-publish.yml
colincasey: Agreed — dropped the author/subject filter (9e86d79); markers post for every main push now.

.github/workflows/build-test-publish.yml
colincasey: Removed (9e86d79). The `HONEYCOMB_API_KEY` secret will be added before merge.

schneems (approved):

PR #1037, opened 2026-08-28
Add Front-end Web CNBs to builder detection

mars:
Supersedes #898.

In preparation for the launch of the Front-end Web CNBs, this adds them to the `heroku-26` builder buildpack detection sequence. This is based on the [approved internal RFC](https://github.com/heroku/rfcs/blob/main/approved/000055-Front-end_Web_Buildpack.md#detect).

The current, recommended builder (`heroku-26`) gains Front-end Web v3.5.0 support for:

- Ember
- create-react-app
- Next.js static exports
- Framework-less sites with `public/index.html`

Each detection group explicitly includes the Node.js and static web server dependencies where required. Framework-specific groups are ordered before the general Node.js group, while Public HTML detection remains last to avoid taking precedence over other language buildpacks.

Vite detection is intentionally excluded because detecting the `vite` dependency alone is too broad and can classify non-static applications as Front-end Web applications.

The `heroku-22` and `heroku-24` builders remain unchanged. The README documents Front-end Web support as specific to `heroku/builder:26`.

## Validation

- Parsed all three builder configurations successfully.
- Verified Front-end Web v3.5.0 buildpacks are declared and ordered only in `builder-26`.
- Verified the pinned Front-end Web image digests against Docker Hub manifests.
- Created and inspected the `builder-26` image locally with `pack builder create`.
- Confirmed `git diff --check` passes.

### DIFF
--- a/.github/workflows/build-test-publish.yml
+++ b/.github/workflows/build-test-publish.yml
@@ -115,10 +115,159 @@ jobs:
             exit 1
           fi
 
+  # The Front-end Web CNBs are only present in builder-26, so this job (unlike the
+  # language `test` job above) doesn't fan out across builder-22/builder-24. There are no
+  # getting-started guides for these frameworks, so each project is scaffolded on the fly
+  # with its standard tooling, then built and served as a static site.
+  test-frontend-web:
+    name: test-frontend-web (${{ matrix.framework.name }}, ${{ matrix.arch }})
+    runs-on: ${{ matrix.arch == 'arm64' && 'ubuntu-24.04-arm' || 'ubuntu-24.04' }}
+    needs: create
+    strategy:
+      fail-fast: false
+      matrix:
+        # Front-end Web is only in builder-26 today, but this is kept as a matrix
+        # dimension (rather than hardcoded) to stay consistent with the create/test
+        # jobs and to make adding a future builder a one-line change.
+        builder: ["builder-26"]
+        arch: ["amd64", "arm64"]
+        # Each framework is scaffolded into `dir` with its standard tooling, then built
+        # and served. `buildpack` is the Front-end Web buildpack asserted to participate in
+        # detection (see the build step); `expected` is a stable string grepped from the
+        # served HTML to confirm the real site was returned (see the test step).
+        framework:
+          - name: create-react-app
+            dir: my-app
+            buildpack: heroku/website-cra
+            expected: "<title>React App</title>"
+          - name: ember
+            dir: ember-quickstart
+            buildpack: heroku/website-ember
+            expected: "<title>EmberQuickstart</title>"
+          - name: nextjs
+            dir: my-app
+            buildpack: heroku/website-nextjs
+            expected: "<title>Create Next App</title>"
+          - name: public-html
+            dir: my-site
+            buildpack: heroku/website-public-html
+            expected: "<title>Heroku Front-end Web</title>"
+    env:
+      FRAMEWORK: ${{ matrix.framework.name }}
+      APP_DIR: ${{ matrix.framework.dir }}
+      EXPECTED: ${{ matrix.framework.expected }}
+      BUILDPACK: ${{ matrix.framework.buildpack }}
+      BUILDER: ${{ matrix.builder }}
+    steps:
+      - name: Install Pack CLI
+        uses: buildpacks/github-actions/setup-pack@812ff667850b930be2217c48e1f980e71bd3e14f # v5.12.3
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
+      - name: Restore Docker images from the cache
+        uses: actions/cache/restore@v5
+        with:
+          fail-on-cache-miss: true
+          key: ${{ github.run_id }}-${{ matrix.builder }}-${{ matrix.arch }}
+          path: images.tar.zst
+        env:
+          SEGMENT_DOWNLOAD_TIMEOUT_MINS: 1
+      - name: Load Docker images into the Docker daemon
+        run: zstd -dc --long=31 images.tar.zst | docker load
+      - name: Scaffold project
+        run: |
+          case "${FRAMEWORK}" in
+            create-react-app)
+              npx --yes create-react-app "${APP_DIR}"
+              ;;
+            ember)
+              # Ember rejects reserved names such as "app", and derives the page <title>
+              # from the app name, so the scaffold name is meaningful here.
+              npm install --global ember-cli
+              ember new "${APP_DIR}" --lang en --skip-git
+              ;;
+            nextjs)
+              npx --yes create-next-app@latest "${APP_DIR}" --yes
+              # A static export is opt-in, so enable it on the scaffolded config. See:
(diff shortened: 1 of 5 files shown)


.github/workflows/build-test-publish.yml
```diff
@@ -115,10 +115,159 @@ jobs:
             exit 1
           fi
 
+  # The Front-end Web CNBs are only present in builder-26, so this job (unlike the
```
edmorley, line 118: In general we've always tried to avoid having stack/builder-specific differences, because it's then yet another thing to document and a potential UX footgun ("why isn't my static app auto-detecting? ... oh it magically behaves differently on different builder versions, and I have the wrong version configured in the Pack CLI default and/or on my Heroku app" etc).

So I would prefer to add it to both `heroku/builder:24` and `heroku/builder:26` at the same time.

Given `heroku/builder:22` is deprecated and EOL at end of April 2027 (and also different in other ways; it's x64 only since backporting arm64 would have been a breaking change), I'm fine with not adding the buildpacks to that builder.

.github/workflows/build-test-publish.yml
```diff
@@ -115,10 +115,159 @@ jobs:
...
+  # language `test` job above) doesn't fan out across builder-22/builder-24. There are no
+  # getting-started guides for these frameworks, so each project is scaffolded on the fly
+  # with its standard tooling, then built and served as a static site.
+  test-frontend-web:
```
edmorley, line 122: I've left some thoughts on testing strategies for what goes in `cnb-builder-images` vs the upstream CNB repo here (not something we necessarily need to fix before this merges):
https://salesforce-internal.slack.com/archives/C01R6FJ738U/p1788518379132859?thread_ts=1787956031.370219&cid=C01R6FJ738U

.github/workflows/build-test-publish.yml
```diff
@@ -115,10 +115,159 @@ jobs:
...
+      FRAMEWORK: ${{ matrix.framework.name }}
+      APP_DIR: ${{ matrix.framework.dir }}
+      EXPECTED: ${{ matrix.framework.expected }}
+      BUILDPACK: ${{ matrix.framework.buildpack }}
+      BUILDER: ${{ matrix.builder }}
+    steps:
+      - name: Install Pack CLI
+        uses: buildpacks/github-actions/setup-pack@812ff667850b930be2217c48e1f980e71bd3e14f # v5.12.3
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
```
edmorley, line 165: This version was very out date, causing GitHub Actions deprecation warnings:

```suggestion
        uses: actions/setup-node@v7
```

.github/workflows/build-test-publish.yml
```diff
@@ -115,10 +115,159 @@ jobs:
...
+            expected: "<title>Heroku Front-end Web</title>"
+    env:
+      FRAMEWORK: ${{ matrix.framework.name }}
+      APP_DIR: ${{ matrix.framework.dir }}
+      EXPECTED: ${{ matrix.framework.expected }}
+      BUILDPACK: ${{ matrix.framework.buildpack }}
+      BUILDER: ${{ matrix.builder }}
+    steps:
+      - name: Install Pack CLI
+        uses: buildpacks/github-actions/setup-pack@812ff667850b930be2217c48e1f980e71bd3e14f # v5.12.3
```
edmorley, line 163: To match the version used elsewhere in this repo:

```suggestion
        uses: buildpacks/github-actions/setup-pack@36416a96be10a6cc7170e7c9c63656d4092f78b2 # v6.1.0
```

.github/workflows/build-test-publish.yml
```diff
@@ -115,10 +115,159 @@ jobs:
...
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
+      - name: Restore Docker images from the cache
+        uses: actions/cache/restore@v5
```
edmorley, line 169:
```suggestion
        uses: actions/cache/restore@v6
```

builder-26/builder.toml
```diff
@@ -57,6 +57,26 @@ mirrors = ["public.ecr.aws/heroku/heroku:26"]
...
+  id = "heroku/static-web-server"
+  uri = "docker://docker.io/heroku/buildpack-static-web-server@sha256:ce6e139ad251da2ab430f0aff405e3d27a3ec052f417827306bfbd5685976dee"
+
+[[buildpacks]]
+  id = "heroku/website-ember"
+  uri = "docker://docker.io/heroku/buildpack-website-ember@sha256:90e1ffb7a92be5f4773f00f2a19ac8cb4166de490d3ce2ac1e48edcbf89bd274"
+
+[[buildpacks]]
+  id = "heroku/website-cra"
+  uri = "docker://docker.io/heroku/buildpack-website-cra@sha256:c55e4a0c20ad74a2b51f7d4ee1f68b1c5b3d28c2ee354f938fece56545569d42"
```
edmorley, line 70: (The rest of the buildpacks are in alphabetical order)

```suggestion
[[buildpacks]]
  id = "heroku/website-cra"
  uri = "docker://docker.io/heroku/buildpack-website-cra@sha256:c55e4a0c20ad74a2b51f7d4ee1f68b1c5b3d28c2ee354f938fece56545569d42"

[[buildpacks]]
  id = "heroku/website-ember"
  uri = "docker://docker.io/heroku/buildpack-website-ember@sha256:90e1ffb7a92be5f4773f00f2a19ac8cb4166de490d3ce2ac1e48edcbf89bd274"
```

.github/workflows/build-test-publish.yml
```diff
@@ -115,10 +115,159 @@ jobs:
...
+          fail-on-cache-miss: true
+          key: ${{ github.run_id }}-${{ matrix.builder }}-${{ matrix.arch }}
+          path: images.tar.zst
+        env:
+          SEGMENT_DOWNLOAD_TIMEOUT_MINS: 1
+      - name: Load Docker images into the Docker daemon
+        run: zstd -dc --long=31 images.tar.zst | docker load
+      - name: Scaffold project
+        run: |
+          case "${FRAMEWORK}" in
```
edmorley, line 180: This `case` statement has no catch-all arm, which means an unmatched framework would silently proceed to the next step. Suggest adding an error message and exit 1 for the catch all.

builder-26/builder.toml
```diff
@@ -156,7 +176,28 @@ mirrors = ["public.ecr.aws/heroku/heroku:26"]
...
+# the Front-end Web buildpacks parse package.json to detect specific dependencies.
+# These groups are intentionally near the end, because projects in other languages often have a package.json.
+[[order]]
+  [[order.group]]
+    id = "heroku/deb-packages"
+    version = "1.0.1"
+    optional = true
+  [[order.group]]
+    id = "heroku/nodejs"
+    version = "5.7.15"
```
edmorley, line 189: The Node.js buildpack has had several updates since this PR was opened - when it's rebased please also check/sync all other versions here and also in the GHA workflow (since other automation and/or Dependbot PRs may have merged since).

